PR #11, opened 2025-03-26
TGC-631 migrate adding value config from forms-runner-v2 repo

alanplatt:


src/server/forms/adding-value.json
```diff
@@ -0,0 +1,1235 @@
...
+      "path": "/score-results",
+      "view": "grants/score-results.html",
+      "next": [],
+      "components": [],
+      "events": {
+        "onLoad": {
+          "type": "http",
+          "options": {
+            "method": "POST",
+            "url": "https://fg-gas-backend.cdpEnvironment.cdp-int.defra.cloud/grants/adding-value/endpoints/calculate-score/invoke"
```
Andrew-Folga, line 302: Are we not going back to calling scoring service directly?

alanplatt: Yes we are, I'll update

swdpcomputing (approved):
LGTM. Missing one line of coverage - 

 server/common/forms/services/form.js  |   93.54 |     90.9 |     100 |   93.54 | 51,70 but it's not necessary for the ticket

PR #30, opened 2025-04-03
RPS-522 - Map state object to GAS

dcapilla:


src/server/land-grants/services/application.service.js
```diff
@@ -0,0 +1,85 @@
...
+  if (stateObj.landParcel && stateObj.actionsObj) {
+    const [sheetId, parcelId] = stateObj?.landParcel?.split('-') ?? []
+    result.actionApplications = []
+
+    Object.entries(stateObj.actionsObj).forEach(([actionCode, actionData]) => {
+      const actionApplication = {
+        code: actionCode
+      }
+
+      if (sheetId) {
```
BugBareDrums, line 58: Wouldn't we want to set this without the check? It won't cause an error and will mean the property is always present, even if it is null/undefined

BugBareDrums: We can chat about this, but I always go for `thing == null` as a check because it will catch null or undefined but nothing else. This would skip sheetId for `0` or empty string

src/server/land-grants/services/application.service.js
BugBareDrums, line 1: Just my opinion but I would have called this file state-to-payload-mapper or something because it is an important function that we will need to locate often

src/server/land-grants/services/application.service.js
dcapilla: yes you are right, makes sense to remove this useless condition. I think I had something else inside, then I refactored some code and left it like that.

src/server/land-grants/services/application.service.js
dcapilla: yes agreed, I think `application.service.js` should contain fetch calls to GAS etc, whereas this mapper should be somewhere else. I will change it.

BugBareDrums (approved):
🐩

PR #33, opened 2025-04-08
Tgc 613 defra id integration

Andrew-Folga:


src/config/nunjucks/context/context.js
```diff
@@ -27,10 +27,21 @@ export function context(request) {
     }
   }
 
+  const session = request?.auth?.isAuthenticated
+    ? await request.server.app.cache.get(request.auth.credentials.sessionId)
+    : {}
+  const auth = {
```
DavidBarker-EE, line 33: If `request` is `null` or it's missing `auth.credentials`, properties like `session.name` will be `undefined`. Not sure if that's intended or not.

src/config/nunjucks/context/context.js
```diff
@@ -18,7 +18,7 @@ let webpackManifest
 /**
  * @param {Request | null} request
  */
-export function context(request) {
+export async function context(request) {
   if (!webpackManifest) {
     try {
       webpackManifest = JSON.parse(readFileSync(manifestPath, 'utf-8'))
```
DavidBarker-EE, line 24: It's an `async` function, so maybe using the async APIs for `readFile` would be better, it'll also improve readability a little imo.

```
  if (!webpackManifest) {
    try {
      const manifestContent = await readFile(manifestPath, 'utf-8')
      webpackManifest = JSON.parse(manifestContent)
    } catch (error) {
      logger.error(`Webpack ${path.basename(manifestPath)} not found:`, error)
    }
  }
```
Not sure if it was intentional to omit the `error` itself in the logger, but it would be quite useful (maybe)? wdyt?

src/plugins/sso.test.js
```diff
@@ -0,0 +1,162 @@
...
+    h = {
+      continue: Symbol('continue'),
+      redirect: jest.fn().mockReturnThis(),
+      takeover: jest.fn().mockReturnValue(Symbol('takeover'))
+    }
+
+    server.ext = jest.fn()
+  })
+
+  afterEach(() => {
```
DavidBarker-EE, line 20: Just a thought about gracefully shutting `server` down after each test in the future, I don't think it's necessary with the tests you've written though.

src/server/auth/get-permissions.test.js
```diff
@@ -0,0 +1,164 @@
...
+    expect(permissionsModule.getPersonId).toHaveBeenCalledWith({ crn, token })
+    expect(permissionsModule.getRolesAndPrivileges).toHaveBeenCalledWith(
+      personId,
+      organisationId,
+      { crn, token }
+    )
+  })
+})
+
+describe('Integration tests for permission functions', () => {
```
DavidBarker-EE, line 145: Maybe separating integration and unit tests into their files would be better, and running them separately if possible.

DavidBarker-EE (dismissed):

src/config/nunjucks/context/context.js
Andrew-Folga: By the time we reach this code the session should theoretically be always present therefore a defensive check is a bit of an overkill. Do you mean I should be adding sth like:

```
  if (!session) {
    // Defensive guard — session should never be missing here
    request.log(['error', 'auth'], 'Authenticated request with missing session data')
    throw new Error('Session data missing despite authenticated request')
  }
```

swdpcomputing (dismissed):

src/config/defra-id.js
```diff
@@ -0,0 +1,56 @@
...
+    env: 'DEFRA_ID_SERVICE_ID'
+  },
+  policy: {
+    doc: 'The Defra Identity policy.',
+    format: String,
+    default: 'default-policy',
+    env: 'DEFRA_ID_POLICY'
+  },
+  redirectUrl: {
+    doc: 'The Defra Identity redirect URL.',
```
alanplatt, line 35: When is this used?

alanplatt (approved):

src/config/defra-id.js
```diff
@@ -0,0 +1,56 @@
+import convict from 'convict'
+
+const config = convict({
+  wellKnownUrl: {
+    doc: 'The Defra Identity well known URL.',
```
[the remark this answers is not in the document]
Andrew-Folga: Clarified on the call.

src/config/defra-id.js
Andrew-Folga: Yes so this is the sign-in-oidc and sign-out-oidc static urls that we workaround by using yar stored dynamic redirect urls in our app. We store in yar before the defra id calls are made and upon getting back to the static redirect configured with DEFRA ID we pull the redirect from yar manager and use it to land back on the original page.

PR #44, opened 2025-04-16
add potential grant funding pages

swdpcomputing:
@ScenarioFactory This solution satisfies the logic in the ticket, and fixes the issue of the 695000 in project-cost not having a comma separator.

However, an important point is that because we are conditionally rendering content rather than having different pages, both of these have the same URL. Is this acceptable?

### DIFF
--- a/src/config/nunjucks/filters/format-currency.js
+++ b/src/config/nunjucks/filters/format-currency.js
@@ -3,11 +3,21 @@
  * @param {Intl.LocalesArgument} locale
  * @param {string} currency
  */
-export function formatCurrency(value, locale = 'en-GB', currency = 'GBP') {
-  const formatter = new Intl.NumberFormat(locale, {
-    style: 'currency',
-    currency
-  })
+export function formatCurrency(
+  value,
+  locale = 'en-GB',
+  currency = 'GBP',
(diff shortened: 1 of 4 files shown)


alexluckett (approved):
> However, an important point is that because we are conditionally rendering content rather than having different pages, both of these have the same URL. Is this acceptable?

As someone outside your team, it seems very reasonable to me. The content is about the potential funding, and the page path is called potential funding. Nice and simple 🙂 

Plus it avoids you needing to duplicate the warning component on line 444 on separate pages.

The duplicated calculation on 454 is not ideal but there's nothing we can do about that for now, given there's no shared state between pages.

Approving as I'd be happy with this if it were my form, but this isn't my repo so you'll probably want another eye from your team.

ScenarioFactory:
> @ScenarioFactory This solution satisfies the logic in the ticket, and fixes the issue of the 695000 in project-cost not having a comma separator.
> 
> However, an important point is that because we are conditionally rendering content rather than having different pages, both of these have the same URL. Is this acceptable?

I think that's fine. The Cap checkers would sometimes use the same URL for subtly different views of the same thing. The elimination pages were the same URLs as the question pages in fact.

PR #50, opened 2025-04-22
Send land grants application to GAS

dcapilla:


src/server/common/services/grant-application.service.js
```diff
@@ -0,0 +1,33 @@
...
+
+  if (!response.ok) {
+    /**
+     * @type {Error & {code?: number}}
+     */
+
+    const data = await response.json()
+    const error = new Error(data.message)
+    error.code = response.status
+    throw error
```
alanplatt, line 29: Mayeb catch and log if there is an error ?

dcapilla: Handling (try/catch) for this is done on the controller. If we can't do anything meaningful with an exception here, it's a good practice to let it propagate.

alanplatt (approved):

PR #57, opened 2025-04-29
Clear cache after submission + basic validations on land parcel and action pages

dcapilla:


bhanu-ukgov (dismissed):

src/server/land-grants/parcel/land-parcel.html
```diff
@@ -11,17 +11,12 @@
...
-              text: error,
-              href: ''
-            }]) %}
-        {% endfor %}
+      {% if selectParcelError %}
         {{ govukErrorSummary({
           titleText: "There is a problem",
-          errorList: errorList
+          errorList: [{
+            text: 'Please select a land parcel from the list'
```
aliuk2012, line 18: Could we not add this error message in the controller? It would make it alot easier to unit test the controller to ensure that the error message is sent to the view if the data is invalid

bhanu-ukgov (approved):

aliuk2012 (approved):

PR #58, opened 2025-04-30
Modify DeclarationPageController to make submission for non land based grants

Andrew-Folga:
For the purpose of this PR I have created a new adding-value grant called `adding-value-v1` (see GAS Swagger for the endpoints https://fg-gas-backend.dev.cdp-int.defra.cloud/documentation#/).

The schema is still being negotiated between GAS and GAE teams but to move the dev forward the schema format was pre assumed in this PR.

Note the answers validation that are sent to GAS is not being hooked up yet as there is no agreement yet as to what should the UX be if the call fails.


swdpcomputing (approved):

src/server/controllers/declaration/gas-answers.schema.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * Validates an object against the applicationAnswersSchema
+ * Only validates types of properties that are present, doesn't require any fields
+ * @param {object} payload - The object to validate
+ * @param {object} options - Joi validation options (optional)
+ * @returns {object} - Joi validation result
+ */
+export function validateApplicationAnswers(
```
ScenarioFactory, line 56: Do we call this function? I'm looking

ScenarioFactory (approved):

src/server/controllers/declaration/gas-answers.schema.js
Andrew-Folga: This is still in the works as we do not have the UX refined yet in case errors occur.

PR #60, opened 2025-04-30
Add new Defra Accounts Bar component and update page headings

aliuk2012:
## Background

To bring the UI closer inline with what is shown in the prototype

## What has changed?
I created a new component called `Defra Account Bar` which displays the current users name, business name and SBI. Its a similar pattern used on Single Front Door but with improved markup (I.e not using `h3` for the larger/bolder text). The component will only display is username and business name are passed to it. 

- Display new Defra Account Bar in the service using fake username/business name which can later be updated from Defra ID
- Updated the serviceName from `grants-ui` to `Manage land-based actions` 
- Added the service navigation component which is the new pattern that GOV.UK will be moving to instead of displaying the name in the Header
- Add simple phase banner for the moment

## Before - Desktop
image

## Before - Mobile
image

## After - Desktop
image

## After - Mobile
image

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -59,7 +59,7 @@ export const config = convict({
   serviceName: {
     doc: 'Applications Service Name',
     format: String,
-    default: 'grants-ui'
+    default: 'Manage land-based actions'
   },
   root: {
     doc: 'Project root',
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -56,6 +56,11 @@ export const config = convict({
     default: oneWeekMs,
     env: 'STATIC_CACHE_TIMEOUT'
   },
+  gitRepositoryName: {
+    doc: 'The name of the git repository which is used for logging events',
+    format: String,
+    default: 'grants-ui'
+  },
   serviceName: {
     doc: 'Applications Service Name',
     format: String,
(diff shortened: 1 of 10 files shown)


src/config/config.js
```diff
@@ -59,7 +59,7 @@ export const config = convict({
   serviceName: {
     doc: 'Applications Service Name',
     format: String,
-    default: 'grants-ui'
+    default: 'Manage land-based actions'
```
dcapilla, line 62: This is also used on the logger options, does this have any implications when writing logs? As we probably want to keep `grants-ui` there.

src/server/common/components/account-bar/template.njk
```diff
@@ -0,0 +1,19 @@
...
+        </div>
+
+        <div class="govuk-grid-column-one-third-from-desktop govuk-!-margin-bottom-2 defra-account-bar__headings defra-account-bar__headings--right-aligned">
+          <span class="govuk-visually-hidden">Signed in as </span>
+          {{ params.userName }}
+        </div>
+      </div>
+    </div>
+  </div>
+{% endif %}
```
dcapilla, line 18: We've gone with no sign out link?

aliuk2012: For the moment, I have agreed with @theboypayne	to leave out the signout link until we have more information about the signout link as at the moment this would be a guess at whether its the correct positioning

src/config/config.js
aliuk2012: I added this commit that creates a new config var for logging [`3ba069a` (#60)](https://github.com/DEFRA/grants-ui/pull/60/commits/3ba069ad788c6afe04e780890a57d7111f0bb0c5)

dcapilla (approved):

PR #66, opened 2025-05-02
Use custom layout for DXT form pages

alexluckett:
Creates a new `layouts/form.njk`, based off our base `layouts/page.njk` template. This `form.njk` template adds any required alterations for a form to be presented.

This PR is not quite ready to be merged, I'm having one issue with the initialisation of a YesNo component on a form - just trying to work that out. It's completely unrelated to this change, not quite sure what's going on!

There's a dependency on the next version of the plugin, which hasn't been published yet. See: https://github.com/DEFRA/forms-engine-plugin/pull/60

Old:
image

New:
image

### DIFF
--- a/src/server/common/forms/definitions/adding-value.json
+++ b/src/server/common/forms/definitions/adding-value.json
@@ -457,7 +457,7 @@
         {
           "type": "YesNoField",
           "name": "remainingCostsYesNoField",
-          "title": " ",
+          "title": "Can you pay the remaining costs?",
           "options": {
             "customValidationMessages": {
               "any.required": "Select yes if you can pay the remaining costs"
--- a/src/server/common/templates/layouts/form.njk
+++ b/src/server/common/templates/layouts/form.njk
@@ -1,0 +1,14 @@
+{% extends "layouts/page.njk" %}
+
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+
+{% block pageTitle %}
+  {{ "Error: " if errors | length }}{{ pageTitle | evaluate }} {{ serviceName }}
+{% endblock %}
+
+{% block beforeContent %}
+  {{ super() }}
+  {% if backLink %}
+    {{ govukBackLink(backLink) }}
+  {% endif %}
+{% endblock %}
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -109,7 +109,7 @@ const registerFormsPlugin = async (server) => {
         formatCurrency
       },
       nunjucks: {
-        baseLayoutPath: 'layouts/form.njk',
+        baseLayoutPath: 'layouts/dxt-form.njk',
         paths: grantsUiPaths
       },
       viewContext: context,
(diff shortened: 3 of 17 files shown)


src/server/common/templates/layouts/form.njk
```diff
@@ -0,0 +1,14 @@
+{% extends "layouts/page.njk" %}
+
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+
+{% block pageTitle %}
+  {{ "Error: " if errors | length }}{{ pageTitle | evaluate }} {{ serviceName }}
```
aliuk2012, line 6:
```suggestion
  {{ "Error: " if errors | length }}{{ pageTitle | evaluate }} | {{ serviceName }}
```

More inline with https://github.com/DEFRA/grants-ui/blob/f6c14ccefb20c20ffb2941dcd40c712476162783/src/server/common/templates/layouts/page.njk#L46C18-L46C21

@dcapilla do you think it might be worth adding the error prefix to our normal layout as well?  I know this is added for accessibility reasons https://design-system.service.gov.uk/patterns/validation/#how-to-tell-the-user-about-validation-errors

src/server/common/templates/layouts/form.njk
aliuk2012, line 1: I think we should rename this to `dxt-form.njk` so its clear this is being used for dxt form page

src/server/common/templates/layouts/form.njk
dcapilla: yes @aliuk2012 - it is a good idea to prefix errors with this as part of our journey, so I am assuming we still have to add it to our custom layout which is used on those custom pages we own.

src/server/common/forms/definitions/adding-value.json
```diff
@@ -457,7 +457,7 @@
         {
           "type": "YesNoField",
           "name": "remainingCostsYesNoField",
-          "title": " ",
+          "title": "Can you pay the remaining costs?",
```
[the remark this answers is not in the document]
dcapilla: I think that's coming from the existing `remainingCostsYesNoField` field  which says "Can you pay the remaining costs of £x?" and formatted a bit to look as a title.

src/server/index.js
```diff
@@ -104,6 +108,11 @@ const registerFormsPlugin = async (server) => {
       filters: {
         formatCurrency
       },
+      nunjucks: {
+        baseLayoutPath: 'layouts/dxt-dxt-form.njk',
```
alexluckett, line 112: typo

Should be `dxt-form.njk`

src/server/common/forms/definitions/adding-value.json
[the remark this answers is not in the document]
alexluckett: Feel free to adjust this, I've got no preference. It just can't be blank like it was before.

As long as there's only one component on the screen it will use the page title, not the the component title, so this doesn't matter too much. It will be used for error messages and summary screens though, so eiter keep the copy I added or use the same value as the page title for consistency.

src/server/common/templates/layouts/form.njk
alexluckett: Good shout

aliuk2012 (approved):

PR #78, opened 2025-05-12
Use GAS endpoint instead of land service for action validation

dcapilla:
At the moment, our UI is calling Land Grants API to validate the actions and quantities submitted by a user, using:

`POST - https://land-grants-api/actions/validate`

This PR is to change that and submit the validation through GAS instead, which will then call our Land Grants API. So our UI will call:

`POST - https://gas-api/grants/frps-private-beta/actions/validate-land-parcel-actions/invoke`


src/server/common/services/grant-application.service.js
```diff
@@ -15,46 +15,69 @@ class GrantApplicationServiceApiError extends Error {
...
+    })
 
     if (!response.ok) {
       const errorText = await response.text()
       throw new GrantApplicationServiceApiError(
         `${response.status} ${response.statusText}`,
         response.status,
         errorText,
-        code
+        grantCode
```
aliuk2012, line 41: This is clearer than the previous `code`

aliuk2012 (approved):

aliuk2012:
Might be worth adding some description to the PR for more context to others in the future.

PR #79, opened 2025-05-13
Update DXT custom form layout to use form name

aliuk2012:
## Issue

DXT form pages were not surfacing the form name but instead displaying the `serviceName`. This fixes the initial issue #66  introduced but we still need to flesh out a bit more on what `grants-ui` servicename will be

## Before
image

## After

image


alexluckett (approved):
Approving from a DXT perspective - looks good. Just confirming we set `name` and `serviceUrl` matcing the form name and form path.

As discussed with Al, in a future iteration we might look at renaming the properties on our end to something like `formName` and `formUrl`. To be discussed when we have clarity on requirements.

dcapilla (approved):

PR #85, opened 2025-05-19
Reinstate GAS submission schema in JSON format and load grant submission validators at startup

Andrew-Folga:


src/server/common/forms/schemas/adding-value-submission-schema.json
```diff
@@ -0,0 +1,89 @@
+{
```
ScenarioFactory, line 1: So we've gone back to a JSON definition of the grant application for the most part?

ScenarioFactory (approved):

src/server/common/forms/schemas/adding-value-submission-schema.json
Andrew-Folga: JSON is only submission schema as it is useless in yaml format.

src/server/common/forms/gas/gas-answers.schema.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/**
+ * Validates an object against the applicationAnswersSchema
+ * Only validates types of properties that are present, doesn't require any fields
+ * @param {object} payload - The object to validate
+ * @param {string} grantCode - The object to validate
+ * @returns {object} - Joi validation result
+ */
+export function validateApplicationAnswers(payload, grantCode) {
```
ScenarioFactory, line 50: Where do we call this method, and what happens if this fails in the UI?

ScenarioFactory: Okay you just answered that question in the stand-up

Andrew-Folga: We do not call it anywhere yet. We will have a follow up story once we decide the UX.

PR #106, opened 2025-06-04
Fix backlist back refresh

swdpcomputing:
Fix tasklist back button persistence and enable Docker hot-reloading

Problems:

  1. Back button disappearing on page refresh - Users lost the "Back to task list" button when refreshing pages, breaking navigation flow
  2. Docker hot-reloading not working - Development workflow required manual container rebuilds for code changes

Solutions:

  1. Back button persistence: Fixed session logic to maintain back button across page refreshes while properly cleaning up when users navigate away
  2. Docker hot-reloading: Set up proper development Docker configuration with volume mounts and file watching


compose.dev.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.4'
```
alanplatt, line 1: the docker-compose file is only used for local dev so we dont need this file

Dockerfile.dev
```diff
@@ -0,0 +1,22 @@
+ARG PARENT_VERSION=latest-22
```
alanplatt, line 1: this should be incorporated into the main docker file and multistage builds used, just use two targets, a default and a dev one see https://docs.docker.com/build/building/multi-stage/

alanplatt (approved):

alanplatt (approved):

PR #110, opened 2025-06-06
Mocking DAL based on env variable

dcapilla:
At the moment, our application queries a DAL url on a real environment which is defined on the CV_API_ENDPOINT environment variable.

This is to add the ability to bypass calling the consolidated DAL API endpoint and serve static data instead.

CV_API_MOCK_ENABLED=true

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -102,7 +102,7 @@ export const config = convict({
     }
   },
   consolidatedView: {
-    mockEnabled: {
+    mockDALEnabled: {
       doc: 'Consolidated View API mock enabled',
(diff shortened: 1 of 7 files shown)


src/config/config.js
```diff
@@ -102,6 +102,12 @@ export const config = convict({
     }
   },
   consolidatedView: {
+    mockEnabled: {
```
aliuk2012, line 105: I think this should be more descriptive so its obvious what its being used eg. `mockDALEnabled` or something along those lines

aliuk2012 (approved):

PR #116, opened 2025-06-09
add v2 to yamls

swdpcomputing:


DavidBarker-EE (commented):
Only the top level `adding-value.yaml` needs this, adding it to all these may have some human readability value but no value for the code itself. `example-grant` should also have `v2` added.

PR #128, opened 2025-06-13
RPS-695 - Get total parcel size area from land grants API & minor template changes

dcapilla:
At the moment, the frontend gets total parcel size area from DAL. We want to change that behaviour and get this data from our Land Grants API. This PR makes that specific change, while adding some file restructuring.

As part of this PR, we are also adding minor changes to the template to match the prototype:

image

### DIFF
--- a/src/server/common/forms/definitions/adding-value/agent-details.yaml
+++ b/src/server/common/forms/definitions/adding-value/agent-details.yaml
@@ -90,7 +90,7 @@ pages:
         title: Address
         name: agentBusinessAddress
     condition: agentDetailsSelected
-  - title: Summary
+  - title: "Check your answers: Agent"
     path: /summary
     controller: SectionEndController
 lists: []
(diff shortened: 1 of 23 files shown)


src/server/land-grants/views/select-land-parcel.html
```diff
@@ -28,7 +28,7 @@
           {% for parcel in parcels %}
             {% set radioItems = radioItems.concat([{
               value: parcel.sheetId + "-" + parcel.parcelId,
-              text: parcel.sheetId + "-" + parcel.parcelId + " (Total size: " + parcel.area + " ha.)"
+              text: parcel.sheetId + "-" + parcel.parcelId + " (Total size: " + parcel.area.value  + " " + parcel.area.unit + ")"
```
aliuk2012, line 31: It might be worth doing this calculation in the controller not in the view. We can add some simple unit tests around this to make sure its coming out in the format we want.

dcapilla: Yes, I agree - I remember having this conversation last week

aliuk2012 (dismissed):

aliuk2012 (approved):

PR #129, opened 2025-06-13
fix taslklist test issues

swdpcomputing:
Update tasklist banner and section headings

  - Change Adding Value tasklist banner from "Manage land-based actions" to "Adding Value Tasklist grant"
  - Fix section end pages to display section titles (e.g. "Business Status") as h1 instead of h2
  - Update redirect URL from /adding-value-tasklist to /adding-value-tasklist/tasklist
  - Fix related tests to match updated URLs and service name

### DIFF
--- a/src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
+++ b/src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
@@ -4,29 +4,59 @@ import {
   TaskListStatus
 } from '../../common/constants/tasklist-status-components.js'
 
+export const SECTIONS = {
+  BUSINESS_STATUS: 'business-status',
+  PROJECT_PREPARATION: 'project-preparation',
+  FACILITIES: 'facilities',
+  COSTS: 'costs',
+  PRODUCE_PROCESSED: 'produce-processed',
+  PROJECT_IMPACT: 'project-impact',
+  MANUAL_LABOUR_AMOUNT: 'manual-labour-amount',
+  FUTURE_CUSTOMERS: 'future-customers',
+  COLLABORATION: 'collaboration',
+  ENVIRONMENTAL_IMPACT: 'environmental-impact',
+  BUSINESS_DETAILS: 'business-details',
+  WHO_IS_APPLYING: 'who-is-applying',
+  AGENT_DETAILS: 'agent-details',
+  APPLICANT_DETAILS: 'applicant-details',
+  CHECK_DETAILS: 'check-details',
+  SCORE_RESULTS: 'score-results',
+  DECLARATION: 'declaration'
+}
+
+export const GRANT_APPLICANT_TYPES = {
+  APPLYING_A1: 'applying-A1',
+  APPLYING_A2: 'applying-A2'
+}
+
+export const ROLES = {
+  APPLICANT: 'applicant',
+  AGENT: 'agent'
+}
+
 const scorePages = [
-  'business-status',
-  'project-preparation',
-  'facilities',
-  'costs',
-  'produce-processed',
-  'project-impact',
-  'manual-labour-amount',
-  'future-customers',
-  'collaboration',
-  'environmental-impact'
+  SECTIONS.BUSINESS_STATUS,
+  SECTIONS.PROJECT_PREPARATION,
+  SECTIONS.FACILITIES,
+  SECTIONS.COSTS,
+  SECTIONS.PRODUCE_PROCESSED,
+  SECTIONS.PROJECT_IMPACT,
+  SECTIONS.MANUAL_LABOUR_AMOUNT,
+  SECTIONS.FUTURE_CUSTOMERS,
+  SECTIONS.COLLABORATION,
+  SECTIONS.ENVIRONMENTAL_IMPACT
 ]
 
 const checkPages = [
   ...scorePages,
-  // 'score-results', // ADD THIS BACK IN WHEN SCORING PAGE WORKS
-  'business-details',
-  'who-is-applying',
-  'agent-details',
-  'applicant-details'
+  // SECTIONS.SCORE_RESULTS, // ADD THIS BACK IN WHEN SCORING PAGE WORKS
+  SECTIONS.BUSINESS_DETAILS,
+  SECTIONS.WHO_IS_APPLYING,
+  SECTIONS.AGENT_DETAILS,
+  SECTIONS.APPLICANT_DETAILS
 ]
 
-const CHECK_DETAILS = 'check-details'
+const CHECK_DETAILS = SECTIONS.CHECK_DETAILS
 
 const declarationPages = [...checkPages, CHECK_DETAILS]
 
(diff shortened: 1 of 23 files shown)


ScenarioFactory (approved):

src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
```diff
@@ -42,17 +42,57 @@ export const addingValueTasklist = {
     register(server) {
       server.route({
         method: 'GET',
-        path: '/adding-value-tasklist',
+        path: '/adding-value-tasklist/tasklist',
```
alanplatt, line 45: this needs to be configurable

src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
alanplatt, line 1: we need to make the filename generic as this feature will be needed by other grants

src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
```diff
@@ -42,17 +42,57 @@ export const addingValueTasklist = {
...
         handler: async (request, h) => {
           const data = (await server.app.cacheTemp.get(request.yar.id)) || {}
           const pageStatuses = determineStatuses(request, data)
           const modelWithStatuses = {
             pageHeading: 'Apply for adding value grant',
```
alanplatt, line 80: should be from config

src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
```diff
@@ -42,17 +42,57 @@ export const addingValueTasklist = {
...
-            sections: applyStatuses(addingValueModel, pageStatuses)
+            sections: applyStatuses(addingValueModel, pageStatuses),
+            serviceName: 'Adding Value Tasklist grant'
```
alanplatt, line 82: config

src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
```diff
@@ -42,40 +72,83 @@ export const addingValueTasklist = {
...
           }
           return h.view('views/adding-value-tasklist-page', modelWithStatuses)
```
alanplatt, line 84: should be a generic name without adding-value

src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
```diff
@@ -42,40 +72,83 @@ export const addingValueTasklist = {
...
         }
       })
+
+      const sections = [
```
alanplatt, line 88: sections should be detailed in config yaml

src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
```diff
@@ -42,40 +72,83 @@ export const addingValueTasklist = {
...
+        SECTIONS.WHO_IS_APPLYING,
+        SECTIONS.AGENT_DETAILS,
+        SECTIONS.APPLICANT_DETAILS,
+        SECTIONS.CHECK_DETAILS
+      ]
+
+      sections.forEach((section) => {
+        server.route({
+          method: 'GET',
+          path: `/adding-value-tasklist/${section}`,
```
alanplatt, line 109: generic slug name from config like dxt grants.

src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
```diff
@@ -42,40 +72,83 @@ export const addingValueTasklist = {
...
+          handler: (request, h) => {
+            const queryString = request.url.search || ''
+            return h.redirect(`/${section}${queryString}`)
+          }
+        })
+
+        server.route({
+          method: 'GET',
+          path: `/adding-value-tasklist/${section}/{path*}`,
```
alanplatt, line 118: as above

src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
```diff
@@ -42,40 +72,83 @@ export const addingValueTasklist = {
...
-    collaboration: TaskListStatus.NOT_YET_STARTED,
-    'environmental-impact': TaskListStatus.NOT_YET_STARTED,
-    'score-results': TaskListStatus.CANNOT_START_YET,
-    'business-details': TaskListStatus.NOT_YET_STARTED,
-    'who-is-applying': TaskListStatus.NOT_YET_STARTED,
-    'agent-details': agentOrApplicantCondition(data, 'agent'),
-    'applicant-details': agentOrApplicantCondition(data, 'applicant'),
-    'check-details': TaskListStatus.CANNOT_START_YET,
-    declaration: TaskListStatus.CANNOT_START_YET
+    [SECTIONS.BUSINESS_STATUS]: TaskListStatus.NOT_YET_STARTED,
```
alanplatt, line 132: sections from configurable from config

alanplatt (changes requested):

ScenarioFactory (dismissed):

alanplatt (dismissed):

src/server/views/section-end-summary.html
```diff
@@ -10,14 +10,10 @@
...
     <div class="govuk-grid-column-two-thirds">
-      {% if pageHeading %}
-        <h1 class="govuk-heading-l">{{ pageHeading | evaluate }}</h1>
-      {% endif %}
-
       {% if checkAnswers %}
         {% for section in checkAnswers %}
           {% if section.title and section.title.text %}
-            <h2 class="govuk-heading-m">{{ section.title.text }}</h2>
+            <h1 class="govuk-heading-l">{{ section.title.text }}</h1>
```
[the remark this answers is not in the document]
ScenarioFactory: If it's the section-end-summary.. why would it have multiple sections in checkAnswers? Can it show more than 1 section in a summary?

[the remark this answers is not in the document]
aliuk2012: `{% for section in checkAnswers %}` made me think that there could end up with more than one `h1` 

Happy to catchup on Monday if its easier to pair on it.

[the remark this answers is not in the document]
swdpcomputing: Indeed. Must have had Friday afternoon brain! Fixed, and added some guards.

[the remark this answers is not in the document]
swdpcomputing: Fixed, and added some guards. Please let me know if requires further attention.

ScenarioFactory (approved):

src/server/views/section-end-summary.html
```diff
@@ -10,26 +10,20 @@
...
       {% if checkAnswers %}
-        {% for section in checkAnswers %}
-          {% if section.title and section.title.text %}
-            <h2 class="govuk-heading-m">{{ section.title.text }}</h2>
-          {% endif %}
-          {% if section.summaryList %}
-            {{ govukSummaryList(section.summaryList) }}
-          {% endif %}
-        {% endfor %}
+        {% if checkAnswers.title and checkAnswers.title.text %}
```
aliuk2012, line 14: I'm still dont think this is correct, if the this condition is false then this page will not have a `h1`

swdpcomputing: I mean, this should never happen. Do you think this should be a breaking error to go to a error page, or a default title?

src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
```diff
@@ -42,40 +72,83 @@ export const addingValueTasklist = {
     register(server) {
       server.route({
         method: 'GET',
-        path: '/adding-value-tasklist',
+        path: '/adding-value-tasklist/tasklist',
```
dcapilla, line 75: can we move this routing out of the controller, please?

src/server/controllers/adding-value-tasklist/adding-value-tasklist-controller.js
dcapilla, line 88: Ideally this controller is generic so Future RPS can use it too with completely different sections. We can work on a way we extend this generic controller and the consumer sets the sections that we want to render, perhaps something like this?

```
export default class FutureRpsTaskListController extends TaskListController {
    constructor() {
      // set sections here
      super.constructor()
    }
}
```

aliuk2012 (approved):
That looks much better, thanks for addressing my comments

PR #136, opened 2025-06-16
Update "Check Your Land Actions" page to support one or more parcels

aliuk2012:
## Background

Initially we only supported one parcel with one or more actions per application. This PR now allows us to storace one or more parcels with one or more actions per applictions


src/server/land-grants/controllers/land-actions-page.controller.js
```diff
@@ -88,10 +88,18 @@ export default class LandActionsPageController extends QuestionPageController {
...
       // Create updated state with the new action data
+      const landParcelID = `${sheetId}-${parcelId}`
       const newState = {
         ...state,
         actions: this.transformActionsForView(actionsObj),
-        actionsObj
+        landParcels: {
+          ...state.landParcels, // Spread existing land parcels
+          [landParcelID]: {
+            id: landParcelID,
```
dcapilla, line 98: do we need this? As it is already part of the data as the key.

dcapilla (approved):

PR #144, opened 2025-06-19
Change "Select actions" page form tabular form to conditionally revealed content

aliuk2012:
### Before
image

### After (Actions collapsed)
image

### After (Actions expanded)
image

### DIFF
--- a/nodemon.json
+++ b/nodemon.json
@@ -4,7 +4,7 @@
     "NODE_ENV": "development"
   },
   "ignore": ["**/*.test.*", "./src/server/common/test-helpers", "./src/client"],
-  "ext": "cjs,js,json",
+  "ext": "cjs,js,json, html, njk",
   "signal": "SIGINT",
   "legacyWatch": true,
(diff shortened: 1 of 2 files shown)


src/server/land-grants/controllers/land-actions-page.controller.js
```diff
@@ -176,11 +176,33 @@ export default class LandActionsPageController extends QuestionPageController {
...
+        })
+      }
+
       // Build the view model exactly as in the original code
       const viewModel = {
         ...this.getViewModel(request, context),
         ...state,
         selectedLandParcelSummary: this.getSelectedLandParcelData(context),
+        selectedActions,
+        selectedActionsQuantities,
```
dcapilla, line 205: maybe we could have this as a single `selectedActions` object? where the key is the selected action and the value is the quantity?

Happy to leave it as is if changing it makes rendering & hydration hard.

dcapilla (approved):

PR #150, opened 2025-06-24
FRPSSFI-33: Improve error message styling

aliuk2012:
### Changes

- Plays back the original data the user submitted before the validations failed
- Splits up the Generic "please select and enter in qty" error message into two seperate validations which can be focused to specific quantity field
- Makes the errors copy in Error summary component a link that takes the user to the invalid field
- Actions/qty "not null" is checked and validated by the UI and only Actions/Qty with values are sent to the API to validate. 
- We merge UI and API validation errors together so users do not have to click twice to get both sets of validation messages

## Before

image

### When user has not selected an action and any quantities
image

### When user has selected an action but not provided the quantity for that actions 
image

### When user has a not provided quantity (UI validation) and the quantity for the other selected actions fail (API validation)
image

### DIFF
--- a/src/server/land-grants/controllers/land-actions-page.controller.js
+++ b/src/server/land-grants/controllers/land-actions-page.controller.js
@@ -142,13 +142,11 @@ export default class LandActionsPageController extends QuestionPageController {
 
   validatePayload = async (payload, actionsObj, sheetId, parcelId) => {
     const errors = {}
-    const errorSummary = []
+
     if (!payload.selectedActions || payload.selectedActions.length === 0) {
       errors.selectedActions = {
-        text: 'Please select at least one action',
-        href: '#selectedActions'
+        text: 'Please select at least one action'
       }
-      errorSummary.push(errors.selectedActions)
     }
 
     if (payload?.selectedActions?.length > 0) {
@@ -156,10 +154,8 @@ export default class LandActionsPageController extends QuestionPageController {
       for (const code of [payload.selectedActions].flat()) {
         if (!payload[`${this.quantityPrefix}${code}`]) {
           errors[code] = {
-            text: `Please provide a quantity for ${code}`,
-            href: `#${this.quantityPrefix}${code}`
+            text: `Please provide a quantity for ${code}`
           }
-          errorSummary.push(errors[code])
         }
       }
     }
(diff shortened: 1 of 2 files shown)


src/server/land-grants/controllers/land-actions-page.controller.js
```diff
@@ -40,6 +40,12 @@ export default class LandActionsPageController extends QuestionPageController {
     return areas
   }
 
+  extractActionsQuantitiesFromPayload(payload) {
```
dcapilla, line 43: perhaps we can merge this method and `extractActionsObjectFromPayload` ?

src/server/land-grants/controllers/land-actions-page.controller.js
```diff
@@ -136,6 +135,78 @@ export default class LandActionsPageController extends QuestionPageController {
...
 
+  validatePayload = async (payload, actionsObj, sheetId, parcelId) => {
+    const errors = {}
+    const errorSummary = []
+    if (!payload.selectedActions || payload.selectedActions.length === 0) {
+      errors.selectedActions = {
+        text: 'Please select at least one action',
+        href: '#selectedActions'
+      }
+      errorSummary.push(errors.selectedActions)
```
dcapilla, line 146: Just a suggestion: you have to maintain `errors` and `errorSummary` every time you want to update errors. Maybe you can merge it into one object and then let the view choose what part of the object wants to render at a specific section?

src/server/land-grants/controllers/land-actions-page.controller.js
```diff
@@ -136,6 +135,78 @@ export default class LandActionsPageController extends QuestionPageController {
...
+          errorSummary.push(errors[code])
+        }
+      }
+    }
+    // Filter actionsObj to only include items with non-null values and ready to be validated by the api
+    const readyForValidationsActionsObj =
+      this.getCompletedFormFieldsForApiValidation(actionsObj)
+
+    if (Object.keys(readyForValidationsActionsObj).length > 0) {
+      const { valid, errorMessages = [] } = await validateLandActions({
```
dcapilla, line 166: I know you haven't changed this, but maybe we should rename `validateLandActions` to `triggerApiActionsValidation` so we know this is the actual API fetch, now that we have a few "validate" functions surrounding it.

src/server/land-grants/controllers/land-actions-page.controller.js
```diff
@@ -136,6 +135,78 @@ export default class LandActionsPageController extends QuestionPageController {
...
+        text: 'Please select at least one action',
+        href: '#selectedActions'
+      }
+      errorSummary.push(errors.selectedActions)
+    }
+
+    if (payload?.selectedActions?.length > 0) {
+      // for each selected action, check if a quantity is provided
+      for (const code of [payload.selectedActions].flat()) {
+        if (!payload[`qty-${code}`]) {
```
dcapilla, line 152:
```suggestion
        if (!payload[`${this.quantityPrefix}-${code}`]) {
```

src/server/land-grants/controllers/land-actions-page.controller.js
aliuk2012: Love this suggestion ❤️

PR #166, opened 2025-07-07
Add apiEndpoint to session cache config. Configure keyGenerator and sessionHydrator that fetches from grants-ui-backend

Andrew-Folga:


src/config/config.js
```diff
@@ -218,6 +218,12 @@ export const config = convict({
         format: Number,
         default: fourHoursMs,
         env: 'SESSION_CACHE_TTL'
+      },
+      apiEndpoint: {
```
Andrew-Folga, line 222: Consider moving to top level

PR #176, opened 2025-07-11
add session rehydration

swdpcomputing:
Session rehydration and SBI authentication

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -10,6 +10,17 @@ import agreements from './agreements.js'
 
 const dirname = path.dirname(fileURLToPath(import.meta.url))
 
+convict.addFormat('url', function (val) {
+  if (!val) {
+    return
+  }
+  try {
+    URL(val)
+  } catch {
+    throw new Error('must be a valid URL')
+  }
+})
+
 const fourHoursMs = 14400000
 const oneWeekMs = 604800000
 
@@ -208,6 +219,12 @@ export const config = convict({
         format: Number,
         default: fourHoursMs,
         env: 'SESSION_CACHE_TTL'
+      },
+      apiEndpoint: {
+        doc: 'Grants UI Backend API endpoint',
+        format: 'url',
+        default: '',
+        env: 'GRANTS_UI_BACKEND_URL'
       }
     },
     cookie: {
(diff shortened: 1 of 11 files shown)


src/server/index.js
```diff
@@ -142,6 +147,158 @@ const registerFormsPlugin = async (server, prefix = '') => {
...
+    serverLogger.debug?.bind(serverLogger) ||
+    serverLogger.info.bind(serverLogger)
+})
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const getIdentity = (_) => {
+  if (process.env.SBI_SELECTOR_ENABLED === 'true') {
+    const sbi = sbiStore.get('sbi')
+    return {
+      userId: `user_${sbi}`,
```
DavidBarker-EE, line 169: This doesn't seem right. The User ID is not the SBI; please use the actual user ID if possible.

swdpcomputing: This will only run on local as we'll turn the envvar for sbi off. I don't think we need sbi for deployed code. So it will then default to the proper userId.

src/config/config.js
```diff
@@ -219,6 +219,12 @@ export const config = convict({
         format: Number,
         default: fourHoursMs,
         env: 'SESSION_CACHE_TTL'
+      },
+      apiEndpoint: {
+        doc: 'Grants UI Backend API endpoint',
+        format: String,
```
alanplatt, line 225: Could this be 'url'  to use Convicts validation?

src/server/index.js
DavidBarker-EE: ah I see

src/server/common/helpers/start-server.test.js
```diff
@@ -77,12 +79,10 @@ describe('#startServer', () => {
...
       )
-      expect(mockHapiLoggerInfo).toHaveBeenNthCalledWith(
-        2,
+      expect(mockHapiLoggerInfo).toHaveBeenCalledWith(
         'Server started successfully'
       )
-      expect(mockHapiLoggerInfo).toHaveBeenNthCalledWith(
-        3,
+      expect(mockHapiLoggerInfo).toHaveBeenCalledWith(
         'Access your frontend on http://localhost:3097'
```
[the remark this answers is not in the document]
swdpcomputing: Resolved.

src/config/config.js
swdpcomputing: Do you want to install convict-format-with-validator? It's an old package... Could do our own validator?

DavidBarker-EE (dismissed):

alanplatt (dismissed):

PR #180, opened 2025-07-14
TGC-750: Example grant with auth - part 1

davidatdefra:
Components and Terminal page

### DIFF
--- a/src/server/common/forms/definitions/example-grant-with-auth.yaml
+++ b/src/server/common/forms/definitions/example-grant-with-auth.yaml
@@ -1,0 +1,549 @@
+engine: V2
+name: Example grant with auth
+metadata:
+  referenceNumberPrefix: EGWA
+  submission:
+    grantCode: example-grant-with-auth
+    submissionSchemaPath: ./schemas/example-grant-with-auth-submission.schema.json
+pages:
+  - title: Check if you can apply for an Example Grant
+    path: '/start'
+    controller: StartPageController
+    components:
+      - name: startInfoOne
+        title: Html
+        type: Html
+        content: |-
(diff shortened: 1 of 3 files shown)


src/server/common/forms/definitions/example-grant-with-auth.yaml
```diff
@@ -0,0 +1,549 @@
...
+        hint: |
+          Use checkboxes to allow selection of one or more options.
+          This example requires at least one selection, but the component can be set to optional.
+        type: CheckboxesField
+        list: projectItemsList
+        options:
+          customValidationMessages:
+            any.required: Select all the items your project needs for now
+
+  - title: What is the estimated cost of the items?
```
swdpcomputing, line 148: This is over-specific for this journey. We don;t want calculations, or specifics like VAT / costs. Just how a number entry works.

src/server/common/forms/definitions/example-grant-with-auth.yaml
```diff
@@ -0,0 +1,549 @@
...
+
+          <ul class="govuk-list govuk-list--bullet">
+          <li>loans</li>
+          <li>overdrafts</li>
+          <li>delinked payments</li></ul>
+
+          <p class='govuk-body'><a href='https://www.gov.uk/guidance/funding-for-farmers' class='govuk-link govuk-link--no-visited-state'>See other grants you may be eligible for</a></p>
+    condition: remainingCostsCondition
+
+  - title: When do you need the grant?
```
swdpcomputing, line 233: I think the validation is conflicting.

alanplatt (approved):

PR #181, opened 2025-07-16
Fix SBI Select list logic

aliuk2012:
The SBI select list was not appearing on the "Find funding for land or farms" forms. This was because this forms start pages page title does not include "find funding" but is called "start page".

I'm still working with DXT forms to get global nunjucks functions supported so we can move this logic out of the views and can be properly unit tested


davidatdefra (approved):
LGTM

PR #194, opened 2025-07-23
State persister for save and return

swdpcomputing:
## Summary

Adds session persistence to DXT to complement session rehydration. This ensures session state is saved not just to cache but also asynchronously to a backend store, enabling Save & Return functionality across devices and sessions.

### Key Changes
- persistStateToApi that is called within DXT's handleSaveAndReturn to persists session state to backend
- Replaces reliance on cacheService.setState for Save & Return
- Improves session durability and supports multi-device use

### Out of scope
This is the first iteration for security scan (note there will be still some work to be added for resilient persister similar to what was done for session purging and for error handling/messaging to the UX when session persisting fails).

### DIFF
--- a/src/server/common/helpers/state/persist-state-helper.js
+++ b/src/server/common/helpers/state/persist-state-helper.js
@@ -12,10 +12,7 @@ export async function persistStateToApi(state, request) {
   const { userId, businessId, grantId } = getCacheKey(request)
   const url = new URL('/state/', GRANTS_UI_BACKEND_ENDPOINT)
 
-  request.logger.info(
-    ['session-persister'],
-    `Persisting state to backend for identity: ${userId}:${businessId}:${grantId}`
-  )
+  request.logger.info(`Persisting state to backend for identity: ${userId}:${businessId}:${grantId}`)
 
   try {
     const response = await fetch(url.href, {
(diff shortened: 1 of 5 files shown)


src/server/index.js
```diff
@@ -222,6 +223,40 @@ async function fetchSavedStateFromApi(request) {
   return json || null
 }
 
+const sessionPersister = async (key, state, request) => {
+  if (!GRANTS_UI_BACKEND_ENDPOINT) {
```
Andrew-Folga, line 227: This is being handled in config.js now in #195

src/server/index.js
```diff
@@ -222,6 +223,40 @@ async function fetchSavedStateFromApi(request) {
...
+    request.logger.warn('Backend not configured - skipping saving data to backend')
+    return
+  }
+
+  const { userId, businessId, grantId } = getIdentity(request)
```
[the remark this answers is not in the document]
swdpcomputing: Done.

src/server/common/helpers/state/persist-state-helper.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+  } catch (err) {
+    request.logger.error({
+      message: `Failed to persist state to API`,
+      err
+    })
+
+    // TODO: See TGC-781
+    // throw err
+  }
```
davidatdefra, line 43: Should have a `return` at the end here as well

Andrew-Folga: When I add back the return I get a linting error:
  43:5  error  Unnecessary return statement  no-useless-return

PR #195, opened 2025-07-23
Implement rehydration in grants-ui - consume DXT session management

Andrew-Folga:
This PR cleans up previous changes done to sbi selector here https://github.com/DEFRA/grants-ui/pull/176

The PR consumes the 2.0.4 version of forms engine plugin that has the new saveAndResume object expecting relevant function to manage DXT form's state. It mainly provides the keyGenerator and sessionHydrator functions that are custom and overwrite the defaults in DXT plugin. Their goal is to provide a way to rehydrate the DXT form's state from a BE application (grants-ui-backend).

It is dependent on DEFRA ID being merged to main which is done under this PR: https://github.com/DEFRA/grants-ui/pull/194

PR #196, opened 2025-07-24
TGC-746: Add session purger and resilient fetch helper

davidatdefra:
Add direct dependencies for @hapi/jwt, @hapi/scooter, accessible-autocomplete, and blankie rather than relying on transitive dependencies from @defra/forms-engine-plugin


package.json
```diff
@@ -53,12 +53,16 @@
     "@hapi/h2o2": "^10.0.4",
     "@hapi/hapi": "^21.4.0",
     "@hapi/inert": "^7.1.0",
+    "@hapi/jwt": "^3.2.0",
```
Andrew-Folga, line 56: Hmm I am not sure we want to overwrite what we actually do not need and is purely required by forms-engine-plugin. Would that be safe? Same with scooter and blankie.

src/server/save-and-return/key-generator.js
```diff
@@ -0,0 +1,35 @@
+import { sbiStore } from '~/src/server/sbi/state.js'
+
+export const keyGenerator = (request) => {
+  const { userId, businessId, grantId } = getIdentity(request)
+  return `${userId}:${businessId}:${grantId}`
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const getIdentity = (_) => {
```
Andrew-Folga, line 9: Same here - this logic needs revising after this PR: https://github.com/DEFRA/grants-ui/pull/195

src/server/common/helpers/resilient-fetch/resilient-fetch.js
```diff
@@ -0,0 +1,32 @@
+import { createLogger } from '~/src/server/common/helpers/logging/logger.js'
+
+export async function resilientFetch(url, options = {}) {
+  const { retries = 0, timeout = 5000, retryDelay = 1000, ...fetchOptions } = options
```
Andrew-Folga, line 4: Nice approach - should we make it into config somehow?

package.json
davidatdefra: All of those dependencies I've added are actually used in our code outside of the plugin, so I think we should have them as direct dependencies, e.g. in [auth.js](https://github.com/DEFRA/grants-ui/blob/1c1ad30c2b076f0b5c2c74593b2c20acb7b25bf5/src/plugins/auth.js#L1), [content-security-policy.js](https://github.com/DEFRA/grants-ui/blob/7e5b7ff6906412cef5f7f5b414ec465a99f5d346/src/plugins/content-security-policy.js#L1), etc

src/server/save-and-return/key-generator.js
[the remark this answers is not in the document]
davidatdefra: I'm happy to refactor to use what you've done. We just happened to be doing similar stuff at the same time and chose different naming.

davidatdefra:
Done in another PR

PR #201, opened 2025-07-30
TGC-768: Defra ID feature flag

davidatdefra:
TGC-696: Enforce Defra ID on all journeys


src/server/auth/index.js
```diff
@@ -103,44 +104,72 @@ function setupBellOAuthErrorHandling(server) {
...
+    server.route({
+      method: 'GET',
+      path: '/auth/organisation',
+      options: {
+        auth: 'defra-id'
+      },
+      handler: handleOrganisationRedirect
+    })
+  } else {
+    // When defra-id is disabled, add a catch-all route for '/auth/*' that redirects to '/home' to avoid 500 errors
```
davidatdefra, line 148: Note: this doesn't work for some reason - will fix, but it doesn't really stop anything else working, just gives ugly 500 error when clicking Sign In button

PR #202, opened 2025-07-31
TGC-768: Update .env.example with `DEFRA_ID_ENABLED` feature flag

davidatdefra:


.env.example
```diff
@@ -34,6 +34,7 @@ GAS_API_URL=https://fg-gas-backend.dev.cdp-int.defra.cloud
 GAS_FRPS_GRANT_CODE=frps-private-beta
 
 # Defra Id Authentication
+DEFRA_ID_ENABLED=false
```
Andrew-Folga, line 37: Do we need to add a section in README as well?

Andrew-Folga, line 37: Docker compose as well maybe?

compose/aws.env
```diff
@@ -2,3 +2,23 @@ AWS_REGION=eu-west-2
 AWS_DEFAULT_REGION=eu-west-2
 AWS_ACCESS_KEY_ID=test
 AWS_SECRET_ACCESS_KEY=test
+
```
Andrew-Folga, line 5: Hey @davidatdefra I've had a look and we do not need this aws.env file as it is part of localstack so no changes needed here and also the localstack and mongodb dependencies should be removed from this repo at some point which we will handle as a separate PR>

PR #204, opened 2025-07-31
Add payment details for selected actions

aliuk2012:
Introduced `annualPaymentPence` property to land actions using mocked payment data. Each action has a hardcoded value for the moment.


src/server/land-grants/controllers/__mocks__/unitRatesForActions.js
dcapilla, line 1: The name of this file + exported default variable is causing a sonar issue

aliuk2012: Great spot, fixed now 👍

PR #231, opened 2025-08-13
TGC-796: Create CheckDetailsPageController

davidatdefra:
Fix minor styling issues across the app


src/server/index.js
```diff
@@ -155,7 +157,8 @@ const registerFormsPlugin = async (server, prefix = '') => {
         SectionEndController,
         FlyingPigsSubmissionPageController,
         PotentialFundingController,
-        SummaryPageController
+        SummaryPageController,
+        CheckDetailsPageController
```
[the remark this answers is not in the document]
davidatdefra: I think Check Answers is the better option, but there's already a Land Grants-specific controller with that name but it does a lot of things specific to that journey. I didn't want to clash, so picked something close.

https://github.com/DEFRA/grants-ui/blob/967ed6a8ea3b9fc76a37e4ac9158368f8d3b0aa4/src/server/land-grants/controllers/check-answers-page.controller.js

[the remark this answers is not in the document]
alanplatt: responses? details implies address etc to me

davidatdefra:
Need to update `@defra/forms-engine-plugin` dependency version after PR https://github.com/DEFRA/forms-engine-plugin/pull/173 is merged and the new version is available

src/server/check-responses/check-responses.controller.js
```diff
@@ -0,0 +1,22 @@
+import { SummaryPageController } from '@defra/forms-engine-plugin/controllers/SummaryPageController.js'
+
+export default class CheckResponsesPageController extends SummaryPageController {
```
swdpcomputing, line 3: The title of the PR is CheckDetailsPageController, but this is named CheckResponsesPageController. Sure this doesn't really matter but just pointing it out in case.

swdpcomputing:
In general it feels like there's some confusion / conflation / mismatch / inconsistency between details and responses, but that may be my misunderstanding.

src/server/check-responses/check-responses.controller.js
davidatdefra: I'll blame Alan for that 😅 I started off calling it CheckDetails, but Alan thought Responses would be better. I'll change the PR title.

src/server/confirmation/views/confirmation-page.html
```diff
@@ -5,38 +5,40 @@
...
-        <p class="govuk-body">Before you start the project, you can:</p>
-        <ul class="govuk-list govuk-list--bullet">
-            <li>get quotes from suppliers</li>
-            <li>apply for planning permission</li>
-        </ul>
-        <div class="govuk-body govuk-!-margin-top-9">
-            <p class="govuk-body"><a href="https://example.com/feedback">What do you think of this service?</a></p>
-        </div>
-    </div>
+<div class="govuk-body">
```
aliuk2012, line 8: why is this div needed? usually this class is added to <p>

src/server/check-responses/views/check-responses-page.html
```diff
@@ -0,0 +1,53 @@
...
+        <input type="hidden" name="crumb" value="{{ crumb }}">
+
+        <div class="govuk-button-group">
+          {{ govukButton({
+            text: "Continue",
+            preventDoubleClick: true
+          }) }}
+
+          {% set allowSaveAndReturn = true %}
+          {% if allowSaveAndReturn %}
```
aliuk2012, line 39: Is this condition needed if the variable will always be true?

PR #232, opened 2025-08-14
complete service to service authentication with basic auth

swdpcomputing:
Add to local .env :

GRANTS_UI_BACKEND_AUTH_TOKEN=<64 character alpha-numeric string no capitals>
GRANTS_UI_BACKEND_ENCRYPTION_KEY=<64 character alpha-numeric string>

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -226,6 +226,20 @@ export const config = convict({
         format: String,
         default: '',
         env: 'GRANTS_UI_BACKEND_URL'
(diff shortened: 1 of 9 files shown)


Andrew-Folga (dismissed):
LGTM

alanplatt:
Might be worth failing fast at boot if GRANTS_UI_BACKEND_URL is set but either token or encryption key is missing. thoughts?

PR #245, opened 2025-08-27
Fix DAL request set sbi correctly for land grants

AndrewKeig:


src/server/common/services/consolidated-view/land-data/107245967.json
aliuk2012, line 1: I wasn't expecting to see this file and doesn't see mto be needed to fix the sbi issue

aliuk2012 (approved):

PR #255, opened 2025-09-02
Persist user answers on Continue

Andrew-Folga:
### Description:

This PR introduces StatePersistenceService, which handles saving and retrieving users’ application answers. Application state is now persisted to grants-ui-backend (MongoDB) whenever users click Continue, ensuring progress is not lost between sessions.

Changes:
- Added StatePersistenceService to manage state lifecycle.
- Removed the legacy “Save and Return” button from the UI.
- Deprecated use of Redis for GrantsUI state management.
- Updated helpers and services to support backend-driven persistence.

Acceptance Criteria:

- When a user answers one or more questions and selects Continue, their responses are persisted in MongoDB.
- Previously saved answers are automatically reloaded when a user returns to their application.

Notes:
- MongoDB (via grants-ui-backend) is now the single source of truth for application state.
- Frontend no longer depends on Redis for state management.

Testing:

- Verified persistence to MongoDB occurs on Continue.
- Confirmed previously answered questions load correctly when revisiting.
- Validated removal of the “Save and Return” button.

### DIFF
--- a/src/server/common/services/state-persistence/state-persistence.service.js
+++ b/src/server/common/services/state-persistence/state-persistence.service.js
@@ -22,7 +22,7 @@ export class StatePersistenceService extends CacheService {
    * @returns {Promise<object>} resolved state or empty object
    */
   async getState(request) {
-    const key = this.Key(request)
+    const key = this._Key(request)
     const state = await fetchSavedStateFromApi(key)
     return state ?? {}
   }
@@ -34,7 +34,7 @@ export class StatePersistenceService extends CacheService {
    * @returns {Promise<object>} the persisted state
    */
   async setState(request, state) {
-    const key = this.Key(request)
+    const key = this._Key(request)
     await persistStateToApi(state, key)
     return state
   }
@@ -48,7 +48,7 @@ export class StatePersistenceService extends CacheService {
    * @returns {Promise<{ confirmed?: true }>} Confirmation state
    */
   async getConfirmationState(request) {
-    const key = this.ConfirmationKey(request)
+    const key = this._ConfirmationKey(request)
     const state = await fetchSavedStateFromApi(key)
     return state ?? {}
   }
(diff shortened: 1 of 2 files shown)


src/server/common/helpers/session-cache/backend-catbox-client.js
```diff
@@ -0,0 +1,100 @@
...
+   */
+  async get(key) {
+    logger.debug(`Cache GET - Key: ${JSON.stringify(key)}`)
+    const state = await fetchSavedStateFromApi(key)
+    logger.debug('Fetched State from API:', state)
+
+    return {
+      item: state ?? null,
+      stored: Date.now(), // Current timestamp so we never consider it stale
+      ttl: ONE_YEAR_MS // Set a long TTL since backend manages actual expiry
```
ScenarioFactory, line 38: Does this have to be set at all? Why a year?

src/server/common/helpers/session-cache/backend-catbox-client.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }
+
+  /**
+   * Override the default set method to add logging
+   * @param {string} key
+   * @param {T} value
+   * @param {number} [ttl]
```
ScenarioFactory, line 46: Is this TTL parameter used?

src/server/common/helpers/session-cache/backend-catbox-client.js
```diff
@@ -0,0 +1,100 @@
+import { createLogger } from '../logging/logger.js'
+import { fetchSavedStateFromApi } from '../state/fetch-saved-state-helper.js'
+import { persistStateToApi } from '../state/persist-state-helper.js'
+
+const ONE_YEAR_MS = 365 * 24 * 60 * 60 * 1000
```
alanplatt, line 5: Why not set it not to expire if we're going to handle this in the backend?

src/server/common/helpers/session-cache/backend-catbox-client.js
alanplatt: Why not set it not to expire if we're going to handle this in the backend?

src/server/common/helpers/session-cache/backend-catbox-client.js
Andrew-Folga: yes it is mandatory by catbox provider.

ScenarioFactory (dismissed):

ScenarioFactory (dismissed):

PR #256, opened 2025-09-02
SFIR-296: Setup users profile regardless of what authentication strategy is used

aliuk2012:
This PR changes how we setup a users profile. We now have a `postAuth` hook which will call the same method regardless what the authentication strategy is. 

if we need to pull back the users name, crn, sbi, organisation name etc we can do so by referring to `request.auth.credentials`. in future we might want to flatten the structure and move to something like `request.userProfile`


alanplatt:
Looks good. My only question really is what about colons in Business name? Do we want to worry about this now? Or at least log the issue?

alanplatt (approved):

alanplatt (approved):

PR #257, opened 2025-09-03
SFIR-189: Show action rates

timney:


src/server/land-grants/controllers/select-actions-for-land-parcel-page.controller.js
```diff
@@ -52,9 +65,27 @@ export default class SelectActionsForLandParcelPageController extends QuestionPa
    * @returns {object} - The view model for the page
    */
   getViewModel(request, context) {
+    const assessMoorlandArea = this.availableActions.find((a) => a.code === 'CMOR1')?.availableArea?.value
```
dcapilla, line 68: Checking against a specific action code here feels too verbose, if we added new actions this would start getting very complex.

This is just my opinion: the backend should group these actions somehow, maybe by adding a property to each action to specify its group. That way the frontend can display those groups automatically.

PR #270, opened 2025-09-08
TGC-808: Use FCP Defra ID Stub

davidatdefra:


compose.yml
```diff
@@ -28,12 +28,32 @@ services:
...
+      - cdp-tenant
+    environment:
+      PORT: 3007
+      AUTH_MODE: ${AUTH_MODE:-basic}
+      AUTH_OVERRIDE: ${AUTH_OVERRIDE}
+      AUTH_OVERRIDE_FILE: ${AUTH_OVERRIDE_FILE}
+    ports:
+      - "3007:3007"
+    volumes:
+      - ./example.data.json:/data/example.data.json
```
ScenarioFactory, line 43: Is this just if we want to add a file of users later?

ScenarioFactory (dismissed):

compose.yml
```diff
@@ -56,24 +78,31 @@ services:
...
+      DEFRA_ID_WELL_KNOWN_URL: ${DEFRA_ID_WELL_KNOWN_URL:-}
+      DEFRA_ID_CLIENT_ID: ${DEFRA_ID_CLIENT_ID:-}
+      DEFRA_ID_CLIENT_SECRET: ${DEFRA_ID_CLIENT_SECRET:-}
+      DEFRA_ID_SERVICE_ID: ${DEFRA_ID_SERVICE_ID:-}
+      DEFRA_ID_REDIRECT_URL: ${DEFRA_ID_REDIRECT_URL:-}
+      DEFRA_ID_SIGN_OUT_REDIRECT_URL: ${DEFRA_ID_SIGN_OUT_REDIRECT_URL:-}
+      COOKIE_PASSWORD: ${COOKIE_PASSWORD:-}
+      GRANTS_UI_BACKEND_URL: ${GRANTS_UI_BACKEND_URL:-}
+      GRANTS_UI_BACKEND_AUTH_TOKEN: ${GRANTS_UI_BACKEND_AUTH_TOKEN:-}
+      GRANTS_UI_BACKEND_ENCRYPTION_KEY: ${GRANTS_UI_BACKEND_ENCRYPTION_KEY:-}
```
alanplatt, line 97: Defaulting to nothing makes debugging hard

compose.yml
davidatdefra: Yeah, I tested out how that worked but decided on going with `basic` as the default so any CRN will work.

Not sure what is best/easiest for testing... just allow anything or have a known set? Do we test login failures?

davidatdefra: Could have an "override" file for CI which includes a file with test users, but leave local dev as basic?

compose.yml
davidatdefra: It's defaulting to nothing here, but the `config` properties should all have defaults, so I thought it best to keep the defaults in one place (mostly).

alanplatt (approved):

PR #276, opened 2025-09-09
Add Another Action to an existing land parcel

aliuk2012:
Displays an option for the user to add another action to a land parcel. This option will only be visible if the user hasn't selected CMOR1 and any UPL

### Before
image

### After
image


src/server/land-grants/controllers/land-actions-check-page.controller.js
```diff
@@ -128,12 +128,38 @@ export default class LandActionsCheckPageController extends QuestionPageControll
     })
   }
 
+  buildLandParcelFooterActions = (selectedActions, sheetId, parcelId) => {
+    const uplCodes = ['UPL1', 'UPL2', 'UPL3']
```
dcapilla, line 132: Rather than hardcoding these values here again, we could just hide then link if 1 action for each existing group has been added.

PR #279, opened 2025-09-09
SFIR-348 - Ability to remove & change land actions

dcapilla:
This PR adds the ability to remove & change land actions from the land grants journey:

image

It also shows which actions have been added when selecting actions for a parcel:

image

### DIFF
--- a/src/server/land-grants/controllers/select-actions-for-land-parcel-page.controller.test.js
+++ b/src/server/land-grants/controllers/select-actions-for-land-parcel-page.controller.test.js
@@ -26,7 +26,7 @@ describe('SelectActionsForLandParcelPageController', () => {
       actions: [
         {
           code: 'CMOR1',
-          description: 'CMOR1: Assess moorland and produce a written record',
+          description: 'Assess moorland and produce a written record: CMOR1',
           availableArea: {
             unit: 'ha',
             value: 10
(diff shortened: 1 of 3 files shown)


src/server/land-grants/controllers/land-actions-check-page.controller.js
```diff
@@ -1,131 +1,130 @@
...
+    links.push(
+      `<li class='govuk-summary-list__actions-list-item'><a class='govuk-link' href='select-actions-for-land-parcel?parcelId=${parcelParam}&action=${data.code}'>Change</a><span class="govuk-visually-hidden"> land action for parcel ${parcel}</span></li>`
+    )
+  }
+  links.push(
+    `<li class='govuk-summary-list__actions-list-item'><a class='govuk-link' href='confirm-remove-action?parcel=${parcelParam}&action=${data.code}'>Remove</a><span class="govuk-visually-hidden"> land action for parcel ${parcel}</span></li>`
+  )
+
+  return {
+    html: `<ul class='govuk-summary-list__actions-list govuk-!-text-align-right'>${links.join('')}</ul>`
```
aliuk2012, line 24: I dont think we should be using summary list action styling inside a table.  Let me have a think about this

dcapilla: we're just using the actions-list styling from the summary list here. But feel free to suggest something else 👍

PR #304, opened 2025-09-18
fix journey unauthorised path

swdpcomputing:


src/config/nunjucks/nunjucks.js
```diff
@@ -46,7 +46,7 @@ export const nunjucksConfig = {
       environment: nunjucksEnvironment
     },
     relativeTo: path.resolve(process.cwd()),
-    path: ['src/server/views', 'src/server'],
+    path: ['src/server/views', 'src/server/auth/views', 'src/server'],
```
DavidBarker-EE, line 49: this does beg the question why we're registering this as a view location and why not put the view with the others in `src/server/views`?

aliuk2012: I agree with @DavidBarker-EE

aliuk2012: It might be worth create a proper route etc that grant forms could forward other unauthorised journeys to. For example, I'm working on a ticket that will make sure that a farmer is authorised to apply for a grant on a land parcel that is owned by their SBI. Ideally any bad requests like this would be sent to the same unauthorised page

swdpcomputing: Have moved.

PR #305, opened 2025-09-18
Set default SBI based on config and empty default mocks

dcapilla:


src/plugins/auth.js
```diff
@@ -221,7 +221,10 @@ function createCredentialsProfile(credentials, payload) {
 }
 
 function extractFarmDetails(relationships) {
-  const parts = relationships.split(':')
+  const parts = relationships?.split(':')
+
+  if (!parts) return []
```
aliuk2012: I dont think we should return empty array. Further down we throw an error because lots of different things refer to the relationship information

PR #312, opened 2025-09-22
Restrict farmers to only Add/Edit/Delete/View land actions for land parcels they own

aliuk2012:
Use a list of Land Parcels from DAL to determine whether the select land parcel is own by the users SBI before allowing them to Add/Edit/Delete/View

When a farmer tries to access a land parcel that their SBI does not own then they will see the following page

image

### DIFF
--- a/src/server/land-grants/controllers/auth/land-grants-question-with-auth-check.controller.js
+++ b/src/server/land-grants/controllers/auth/land-grants-question-with-auth-check.controller.js
@@ -1,6 +1,7 @@
 import { QuestionPageController } from '@defra/forms-engine-plugin/controllers/QuestionPageController.js'
 import { statusCodes } from '~/src/server/common/constants/status-codes.js'
 import { fetchParcels } from '~/src/server/land-grants/services/land-grants.service.js'
+import { stringifyParcel } from '~/src/server/land-grants/utils/format-parcel'
 
 export default class LandGrantsQuestionWithAuthCheckController extends QuestionPageController {
   landParcelsForSbi = []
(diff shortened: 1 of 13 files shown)


src/server/land-grants/controllers/land-parcel-page.controller.js
```diff
@@ -27,6 +30,10 @@ export default class LandParcelPageController extends QuestionPageController {
       const payload = request.payload ?? {}
       const { selectedLandParcel, action } = payload
 
+      this.selectedLandParcel = selectedLandParcel
+
+      if (!this.landParcelBelongsToSbi()) { return this.renderUnauthorizedView(request, h) }
+
```
dcapilla, line 36: I wonder, could we even move this check to the parent class? so here we just have something like 

```
const authResult = await this.performGetAuthCheck(request, context, h)
 if (authResult) return authResult
```

and then on the parent controller we have something like

```
performAuthCheck = async (request, context, h) => {
  if (!this.landParcelBelongsToSbi()) {
    return this.renderUnauthorizedView(request, h)
  }
  return null
}
```

I think the less code we have on the child class, the better.

src/server/land-grants/controllers/remove-action-page.controller.js
```diff
@@ -174,6 +177,12 @@ export default class RemoveActionPageController extends QuestionPageController {
       if (!parcel) {
         return this.proceed(request, h, checkSelectedLandActionsPath)
       }
+      const parcels = await fetchParcels(request.auth.credentials.sbi)
```
dcapilla, line 180: maybe this fetch can be moved to the parent too in case it doesn't find any parcels on the current state? I am not sure

src/server/land-grants/controllers/remove-action-page.controller.js
```diff
@@ -193,6 +202,13 @@ export default class RemoveActionPageController extends QuestionPageController {
       const { state } = context
       const payload = request.payload ?? {}
 
+      if (!ownsLandParcel(this.landParcelsForSbi, this.parcel)) {
```
dcapilla, line 205: aren't we checking this already on line 184?

src/server/land-grants/controllers/auth/land-grants-question-with-auth-check.controller.js
```diff
@@ -0,0 +1,26 @@
...
+export default class LandGrantsQuestionWithAuthCheckController extends QuestionPageController {
+  landParcelsForSbi = []
+  selectedLandParcel = ''
+
+  landParcelBelongsToSbi = () => {
+    return this.landParcelsForSbi.includes(this.selectedLandParcel)
+  }
+
+  performAuthCheck = async (request, h) => {
+    const landParcels = (await fetchParcels(request.auth.credentials.sbi)) || []
```
dcapilla, line 14: should we cache the response of this so we don't call fetchParcels every time for the same sbi?

aliuk2012: I did think about caching but thought we should leave that for the moment because it could introduce issues in the future when users are able to switch sbi and we would need to keep track of which sbi they are using at the time.

PR #332, opened 2025-09-30
TGC-840: Application status

davidatdefra:


src/server/common/helpers/state/persist-submission-helper.js
```diff
@@ -0,0 +1,52 @@
...
+      })
+    }
+  } catch (err) {
+    log(LogCodes.SYSTEM.EXTERNAL_API_ERROR, {
+      method: 'POST',
+      endpoint: url.href,
+      referenceNumber: submission.referenceNumber,
+      error: err.message
+    })
+    // TODO: See TGC-781
```
Andrew-Folga, line 49: Does this ticket exist - it did not load for me in JIRA and if so do we need creating a new one? I think this one resulted from the copy of the state persister correct?

src/server/common/helpers/state/update-application-status-helper.js
```diff
@@ -0,0 +1,56 @@
...
+      })
+    }
+  } catch (err) {
+    log(LogCodes.SYSTEM.EXTERNAL_API_ERROR, {
+      method: 'PATCH',
+      endpoint: url.href,
+      identity: key,
+      error: err.message
+    })
+    // TODO: See TGC-781
```
Andrew-Folga, line 53: Does this ticket exist - it did not load for me in JIRA and if so do we need creating a new one? I think this one resulted from the copy of the state persister correct?

davidatdefra: Yes, it's copied from the persist state - I cannot find the ticket either, but probably better to keep all 3 comments the same for easier finding of the instances we need to look at.

src/server/common/helpers/state/update-application-status-helper.js
Andrew-Folga, line 1: Is this a good place to add this? It is more about a status not a state so I would move it into its own `/status` directory.

src/server/common/helpers/state/update-application-status-helper.js
davidatdefra: Added TGC-873 and changed TODO comment to point there

ScenarioFactory (approved):

PR #347, opened 2025-10-03
SFIR-490 - Add ability to clear application state for non-prod environments

dcapilla:
This PR adds a "Clear application state" link to the footer for non production environments:

image

This clears the application state on `grants-ui-backend` based on SBI and current url slug. If no sbi or slug are present, state won't be cleared.


src/server/common/templates/layouts/page.njk
```diff
@@ -76,22 +76,31 @@
...
+      href: "https://www.gov.uk/help/cookies",
+      text: "Cookies"
+    },
+    {
+      href: "https://www.gov.uk/help/accessibility-statement",
+      text: "Accessibility statement"
+    }
+  ] %}
+  
+  {% if cdpEnvironment !== 'prod' %}
```
aliuk2012, line 94: Wouldn't we want to clear the state in local dev too?

aliuk2012 (approved):

PR #365, opened 2025-10-08
add dynamic text to browser tab and ServiceNavigation. Also updated l…

swdpcomputing:
…ands grant url


src/server/confirmation/services/confirmation.service.js
```diff
@@ -85,16 +89,24 @@ export class ConfirmationService {
...
     form = null,
-    slug = null
+    slug = null,
+    formDefinition = null
   }) {
+    // Get serviceName from form definition metadata, fallback to global config default
+    const serviceName = formDefinition?.metadata?.serviceName || 'Manage land-based actions'
+
+    // Get serviceUrl from form definition metadata, fallback to form path or old default
+    const serviceUrl = formDefinition?.metadata?.serviceUrl || (slug ? `/${slug}` : '/find-funding-for-land-or-farms')
```
aliuk2012, line 99: I'm a bit nervous about the naming of `serviceName` as there is already a global for this and it might cause some weird bugs

PR #470, opened 2025-11-20
Update land parcels in the mock dal data

HariQA-commits:
- Update land parcels in the mock dal data.
- Existing land parcels are now not on moorland and can't use for e2e testing and also to add 5 land parcels to each mock sbi instead only one land parcel.


aliuk2012 (approved):

miriyalanarahari:
Not required anymore.Made another pr with these changes.

PR #754, opened 2026-04-07
POC Edit details through SFD

DavidBarker-EE:


src/server/details-page/check-details.controller.js
```diff
@@ -246,13 +247,40 @@ export default class CheckDetailsController extends QuestionPageController {
...
+  getSFDUpdateUrl(request) {
+    const { sbi } = request.auth.credentials
+    const updateUrl = config.get('externalLinks.sfd.updateUrl')
+    if (!updateUrl) {
+      return ''
+    }
+
+    try {
+      const url = new URL(updateUrl)
+      url.searchParams.set('ssoOrgId', sbi)
```
aaroncarroll82: I don't think it's the SBI you want to set here - it's the organisation ID that comes back from the data returned by DAL

aaroncarroll82 (approved):

PR #779, opened 2026-04-09
Save totalHectaresAppliedFor as part of post handler when saving landParcels

Andrew-Folga:


src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
```diff
@@ -131,8 +131,16 @@ export default class CommonSelectLandParcelPageController extends LandGrantsQues
       })
     }
 
+    const parcelMap = new Map(fetchedParcels.map((p) => [`${p.sheetId}-${p.parcelId}`, p]))
```
dcapilla: here you can use the stringifyParcel method  on format-parcel.js

```
const parcelMap = new Map(fetchedParcels.map((p) => [stringifyParcel(p), p]))
```

src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
```diff
@@ -131,8 +131,16 @@ export default class CommonSelectLandParcelPageController extends LandGrantsQues
...
+      return sum + (parcel?.area?.value || 0)
+    }, 0)
+
     await this.mergeState(request, state, {
-      landParcels: selectedParcelIds
+      landParcels: selectedParcelIds,
+      totalHectaresAppliedFor
     })
 
     return this.proceed(request, h, `${this.getNextPath(context)}?selectedParcelIds=${selectedParcelIds.join(',')}`)
```
dcapilla: do we need to add `selectedParcelIds` as part of the url? I don't think we need that as they are saved into the state at this point.

dcapilla (commented):
just a couple of minor comments, other than that looks good!

src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
Andrew-Folga: good point fixed

src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
Andrew-Folga: yep removed

davidatdefra (approved):

PR #786, opened 2026-04-13
Allow land parcels to display on summary page

Andrew-Folga:


src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
```diff
@@ -162,6 +162,7 @@ export default class CommonSelectLandParcelPageController extends LandGrantsQues
 
     await this.mergeState(request, state, {
       landParcels: selectedParcelIds,
+      landParcelsDisplay: selectedParcelIds.join(', '),
```
davidatdefra, line 165: Is this intended to be to display the value in the check your answers page? Because I don't think it will - that page doesn't currently display `landParcelMetadata` or `totalHectaresAppliedFor`

I'm pretty sure it only displays the contents of `relevantState` rather than `state`, and data only gets added there if it has a "real" DXT component on a page afaik

Andrew-Folga: I tested it and it worked. This is not intended for `metadata` or the `hectars` - this is not meant to be displayed there on the summary
This is for landParcelsDisplay only which is handled inside getSummaryViewModel

davidatdefra (dismissed):

src/server/check-responses/check-responses.controller.js
```diff
@@ -30,6 +49,8 @@ export default class CheckResponsesPageController extends SummaryPageController
     const backLink = getTaskPageBackLink(viewModel, pageDef)
     const sectionTitle = this.section?.hideTitle !== true ? this.section?.title : ''
 
+    this.#applyLandParcelsDisplay(viewModel, context?.state?.landParcelsDisplay)
```
dcapilla: I am a bit reluctant about using the field text ('Select land parcels') to match the target row, it seems a bit fragile. Perhaps we could use the state variable name instead?


```suggestion
    const landParcels = context?.state?.landParcels
    if (Array.isArray(landParcels) && landParcels.length) {
      const displayValue = landParcels.join(', ')
      viewModel.details?.forEach((detail, di) => {
        const ii = detail.items?.findIndex((item) => item.name === 'landParcels') ?? -1
        if (ii !== -1) {
          viewModel.checkAnswers[di].summaryList.rows[ii].value = { html: displayValue }
        }
      })
    }
```

Andrew-Folga: I am afraid there is no stable identifier like you say. The viewModel around checkAnswers  details is:

```
...
  {

    "title": {

      "text": "About your woodland"

    },

    "summaryList": {

      "rows": [

        {

          "key": {

            "text": "Select land parcels"

          },

          "value": {

            "classes": "app-prose-scope",

            "html": "Not provided"

          },

          "actions": {

            "items": [

              {

                "href": "/woodland/land-parcels?returnUrl=%2Fwoodland%2Fsummary",

                "text": "Change",

                "classes": "govuk-link--no-visited-state",

                "visuallyHiddenText": "Select land parcels"

              }

            ]

          }

        },
        ...
```

Andrew-Folga: I've looked into it again and as mentioned above there’s a misunderstanding here about what we actually have available in the `summaryList` structure.

There is no stable field identifier coming through from the state into checkAnswers or rows (e.g. no name, id, or fieldKey that maps to landParcels). The only data we have in the rendered summary is:

- key.text (e.g. "Select land parcels")
- value (rendered output like "Not provided")
- actions.visuallyHiddenText

So we cannot reliably use state.landParcels to locate the row directly, because that metadata is not preserved in the view model.

dcapilla: @Andrew-Folga  -  we don't need anything available in `summaryList` for this to work. We just need to edit the value of the row on that list.

In order to do so, viewModel.details is also available in getSummaryViewModel, and details[i].items[j].name maps directly to the YAML field name ('landParcels'). Since checkAnswers is built by a direct .map() over details, the indexes align 1:1, so we can locate the row without any label matching:

In the approach already suggested:

```
viewModel.details?.forEach((detail, di) => {
  const ii = detail.items?.findIndex((item) => item.name === 'landParcels') ?? -1
  if (ii !== -1) {
    viewModel.checkAnswers[di].summaryList.rows[ii].value = { html: displayValue }
  }
})
```

We find by item.name in details, then use the index to patch the corresponding checkAnswers row. No label matching needed.

src/server/check-responses/check-responses.controller.js
```diff
@@ -16,22 +16,16 @@ export default class CheckResponsesPageController extends SummaryPageController
...
-      if (targetRow) {
-        targetRow.value = { text: landParcelsDisplay }
-      }
+  #applyLandParcels(viewModel, landParcels) {
+    if (Array.isArray(landParcels) && landParcels.length) {
+      const displayValue = landParcels.join(', ')
+      viewModel.details?.forEach((detail, di) => {
+        const ii = detail.items?.findIndex((item) => item.name === 'landParcels') ?? -1
+        if (ii !== -1) {
+          detail.items[ii].value = displayValue
```
dcapilla: I don't think you need this line, just need to change the viewModel value

Andrew-Folga: Addressed - thanks

src/server/check-responses/check-responses.controller.js
Andrew-Folga: As agreed we will open a new ticket https://eaflood.atlassian.net/browse/TGC-1220

davidatdefra (approved):

PR #789, opened 2026-04-13
TGC-1174: Update compose with config api and config broker

davidatdefra:


localstack/config-broker/example-grant-with-auth/example-grant-with-auth.yaml
aaroncarroll82, line 1: I noticed this was moved from grants-ui/x.yml-> x.yml . Would it be better to keep it in the grants-ui directory here so it represents where the deployed config broker would be storing this and offering it up from? The files and their locations are definitely up for grabs in terms of how they're named and laid out, but I think likely keeping the UI specific bits of config under a ui directory will make sense once the other parts of config are started to be added alongside

broker.http
```diff
@@ -0,0 +1,29 @@
...
+@slug = example-grant-with-auth
+@semver = 1.0.0
+
+### List All Grants
+# @name GetAllGrants
+GET {{brokerUrl}}/api/allGrants
+
+###
+
+### Get Grant Version Detail
```
aaroncarroll82, line 25: Worth adding the endpoint for the latest version in here, as I assumed that is how the config API would be checking for the config version to use with that endpoint?

PR #796, opened 2026-04-15
Feature/tgc 1196 update woodland v3

swdpcomputing:


src/server/common/forms/definitions/woodland.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
   submission:
     grantCode: woodland
   grantRedirectRules:
+    excludedPaths: ['check-details']
```
davidatdefra: Because `check-details` is the start page, this breaks all the redirect logic for both pre and post submission. This needs to be removed.

I suspect this was done to prevent that redirect from kicking in when using the back link? If so, there is another way™

```
diff --git a/src/server/status/status-helper.js b/src/server/status/status-helper.js
--- a/src/server/status/status-helper.js	(revision Staged)
+++ b/src/server/status/status-helper.js	(date 1776288593036)
@@ -337,9 +337,10 @@
 
   const previousStatus = context.state.applicationStatus
   const grantRedirectRules = request.app.model?.def?.metadata?.grantRedirectRules
+  const isWithinGrantPages = request.headers['sec-fetch-site'] === 'same-origin' // request initiated from within grant pages
 
   // Don't redirect if page is listed in the grant config excludedPaths
-  if (grantRedirectRules?.excludedPaths?.includes(request.params?.path)) {
+  if (grantRedirectRules?.excludedPaths?.includes(request.params?.path) || isWithinGrantPages) {
     return h.continue
   }
``` 

I think if we put this in, we need to remove the `isWithinGrantPages` part of the if statement within `handlePostSubmission`

src/server/common/forms/definitions/woodland.yaml
davidatdefra: This should be `/check-details` as that is now the "start" page

src/server/details-page/views/incorrect-details.html
```diff
@@ -10,31 +10,40 @@
...
 
-      <p class="govuk-body govuk-!-margin-top-6">
-        Once your details have been updated, you can return to this service and continue with your application.
-      </p>
+        <p class="govuk-body govuk-!-margin-top-6">
+          Once your details have been updated, you can return to this service and continue with your application.
+        </p>
+      {% endif %}
 
       {{ govukButton({
```
davidatdefra: The prototype doesn't have any button on this page

src/server/details-page/check-details.controller.js
```diff
@@ -126,33 +124,34 @@ export default class CheckDetailsController extends QuestionPageController {
...
    */
-  async handleDetailsConfirmed(request, context, detailsConfig, h) {
+  async handleDetailsConfirmed(request, context, config, h) {
     const baseViewModel = super.getViewModel(request, context)
 
     try {
-      const { mappedData } = await this.fetchAndProcessData(request, detailsConfig)
-      await this.setState(
-        request,
-        mergeAdditionalAnswers(context.state, {
```
davidatdefra, line 151: Any reason not to use the common helper method?

src/server/common/forms/definitions/woodland.yaml
ScenarioFactory: I did a little pre-emptive work updating the acceptance tests against this branch and I did find the redirection on GAS status was broken, yeah.

src/server/common/forms/definitions/woodland.yaml
```diff
@@ -189,7 +200,30 @@ pages:
...
+        type: Details
+        title: Help with your application
+        content: |
+          <p class="govuk-body">This task list helps you track your progress through the application. Each section contains one or more tasks that must be completed.</p>
+          <p class="govuk-body">Tasks will show as 'Completed' as they are populated. You must complete all sections before you can submit your application.</p>
+          <p class="govuk-body">If you need to change any information you have already provided, you can go back to any completed section and update your answers. The status will update automatically as you progress through the application.</p>
+        options:
+          position: above
+
+  # Start page (kept for backward compatibility — journey starts at /check-details)
```
davidatdefra: Not sure what this means? Backward compatibility? If farmers aren't mean to see this page, we should remove it.

davidatdefra (approved):

PR #797, opened 2026-04-15
TGC-1174: Config broker integration

davidatdefra:


compose.yml
```diff
@@ -8,6 +8,10 @@ services:
         condition: service_healthy
       grants-ui-backend:
         condition: service_started
+      grants-ui-config-api:
+        condition: service_healthy
+      grants-config-broker:
```
aaroncarroll82: Wondering if it's worth removing the indirect dependencies here as depends_on conditions, just to aid clarity and remove redundant repetition? Unless I have misunderstood grants-ui _depends on_ grants-ui-config-api _which depends on_ config-broker

src/server/dev-tools/handlers/clear-application-state.handler.js
```diff
@@ -12,6 +13,14 @@ export async function clearApplicationStateHandler(request, h) {
 
   // we need a slug to clear the persisted state
   if (slug) {
+    if (!request.app.model) {
+      const form = await findFormBySlug(slug)
+      if (form) {
+        const definition = await loadFormDefinition(form)
```
alanplatt: if the cache has expired, the loadFormDefinition will throw and we wont get the clearstate below

davidatdefra: Refactored

alanplatt (approved):

PR #812, opened 2026-04-22
Prevent returnUrl redirect from summary when editing land parcels

Andrew-Folga:
Prevents users being redirected back to the summary page when editing land parcels via “Change”.

Strips `returnUrl` on GET for the parcel selection page so the journey continues normally via `getNextPath`.

This fixes an issue where users could submit more combined hectarage than selected parcels due to unintended navigation back to summary.

### DIFF
--- a/src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
+++ b/src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
@@ -187,7 +187,9 @@ export default class CommonSelectLandParcelPageController extends LandGrantsQues
       additionalAnswers: { totalHectaresAppliedFor }
     })
 
-    return this.proceed(request, h, `${this.getNextPath(context)}`)
+    const { returnUrl: _removed, ...queryWithoutReturnUrl } = request.query ?? {}
+    const requestWithoutReturnUrl = { ...request, query: queryWithoutReturnUrl }
+    return this.proceed(requestWithoutReturnUrl, h, `${this.getNextPath(context)}`)
   }
 }
 
--- a/src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
+++ b/src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
@@ -28,6 +28,16 @@ export default class CommonSelectLandParcelPageController extends LandGrantsQues
     this.minimumAreaHa = config.minimumAreaHa ?? null
   }
 
+  makeGetRouteHandler() {
+    return async (request, context, h) => {
+      if (request.query?.returnUrl) {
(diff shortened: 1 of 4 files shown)


src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
```diff
@@ -28,6 +28,16 @@ export default class CommonSelectLandParcelPageController extends LandGrantsQues
     this.minimumAreaHa = config.minimumAreaHa ?? null
   }
 
+  makeGetRouteHandler() {
+    return async (request, context, h) => {
+      if (request.query?.returnUrl) {
```
dcapilla, line 33: This is fine for now, but will be applied to other journeys using the common land parcel page controller and we might not always want this behaviour. We might need to think of a flexible solution.

Happy to approve as it doesn't affect other journeys now, but one to think about.

dcapilla (dismissed):

src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
```diff
@@ -28,6 +28,16 @@ export default class CommonSelectLandParcelPageController extends LandGrantsQues
     this.minimumAreaHa = config.minimumAreaHa ?? null
   }
 
+  makeGetRouteHandler() {
```
davidatdefra, line 31: I think we can do this without the extra redirect

~Instead, just add `backLink: undefined` to `buildViewModel`~

I haven't tested this though, just a thought from looking at the DXT code

davidatdefra: Ok, actually that did not work on testing, so ignore that.
But... this does work...

Change the return at the end of `handlePost` to remove returnUrl from request.query
```
    const { returnUrl: _removed, ...queryWithoutReturnUrl } = request.query ?? {}
    const requestWithoutReturnUrl = { ...request, query: queryWithoutReturnUrl }
    return this.proceed(requestWithoutReturnUrl, h, `${this.getNextPath(context)}`)
```

src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
Andrew-Folga: Changed to the solution @davidatdefra proposed below hence closing this off.

src/server/land-grants/common/common-select-parcel/common-select-land-parcel-page.controller.js
Andrew-Folga: Thanks and this is addressed now.

DavidBarker-EE (approved):

PR #822, opened 2026-04-24
TGC-1194 - Trying to fix pact flakiness - second attempt

dcapilla:
After we bumped @pact-foundation/pact from 15.0.1 to 16.3.0, the contract tests started failing intermittently on the self-hosted runner. Different tests each time, which made it hard to track down.

We suspect it is due to pact-js 16.x dragging in a new major version of @pact-foundation/pact-core causing this problem.

Changes:

- Pinned @pact-foundation/pact back to 15.0.1
- Added a dependabot ignore rule for >= 16.0.0
- Clean the pact output directory before running tests on the shared runner
- Mocked the retry helper in the contract test files that were missing it

Once pact-core 19.x stabilises, remove the ignore rule and retest.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,8 @@ updates:
     ignore:
       - dependency-name: "stylelint"
         update-types: ["version-update:semver-major"]
+      - dependency-name: "@pact-foundation/pact"
+        versions: [">=16.0.0"]
     groups:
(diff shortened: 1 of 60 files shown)


DavidBarker-EE (dismissed):

Andrew-Folga (approved):
LGTM

PR #826, opened 2026-04-24
TGC-1259: Add hoisting for API based config form, to match direct fil…

aaroncarroll82:
1. Hoist page config to metadata for form definitions loaded via API
2. Use correct compose config so that config API is used to provide example-grant-with-auth definition
  a. Update yaml used in compose setup to come from broker + API to match latest definition
3. Ensure version can be pulled from metadata in controllers not extending PageController in formsPlugin
4. Expose formsService via server.methods to ensure places that need it can actually get to it


dcapilla (dismissed):

src/server/confirmation/config-confirmation.js
```diff
@@ -135,14 +97,37 @@ export const configConfirmation = {
...
         path: '/{slug}/confirmation',
+        options: {
+          pre: [{ method: validateRequestAndFindForm, assign: 'validatedSlugAndForm' }]
+        },
         handler: async (request, h) => {
           try {
-            const validationResult = await validateRequestAndFindForm(request, h)
-            if (validationResult.error) {
-              return validationResult.error
+            const { validationResult: validationError, form, slug } = request.pre.validatedSlugAndForm
```
swdpcomputing: validationError is always falsy, as destructutng validationResult gives undefined, so the handler will always fall to getSessionDataAndState, throwing 500.

src/server/confirmation/config-confirmation.test.js
```diff
@@ -100,22 +101,20 @@ describe('config-confirmation', () => {
       expect(mockH.view).toHaveBeenCalledWith('config-confirmation-page', { test: 'viewModel' })
     })
 
-    test('should return 400 when slug is missing', async () => {
-      mockRequest.params = {}
+    test('should return validation error when pre-handler returns it', async () => {
```
swdpcomputing: I think this will always give a false positive. You're comparing apples for apples with the mock.

src/server/common/helpers/form-verify-and-request-load.test.js
```diff
@@ -0,0 +1,40 @@
...
+  it('should return form and slug and set request app model when inputs valid', async () => {
+    const formWithMetaData = { ...mockForm, metadata: { some: 'metadata' } }
+    findFormBySlug.mockResolvedValue(formWithMetaData)
+    const mockH = mockHapiResponseToolkit()
+    const requestProps = { params: { slug: 'test-slug' }, app: {} }
+    const result = await validateRequestAndFindForm(mockHapiRequest(requestProps), mockH)
+    expect(result.error).toBeUndefined()
+    expect(result.form).toBe(formWithMetaData)
+    expect(result.slug).toBe('test-slug')
+    expect(requestProps.app.model.def.metadata).to.eql({ some: 'metadata' })
```
swdpcomputing: Prefer .toEqual(...) to keep in line with other tests

src/server/dev-tools/handlers/clear-application-state.handler.js
```diff
@@ -50,3 +49,14 @@ export async function clearApplicationStateHandler(request, h) {
 
   return h.redirect(`/${slug}`)
 }
+
+const loadFormAndSetOnRequestModel = async (form, request) => {
+  const getFormService = request.server.methods.getFormService
+
+  if (typeof getFormService !== 'function') {
```
swdpcomputing: In prod this can;t throw as getFormService is registered unconditionally. As executing during startup, it will always be a function.

src/server/common/helpers/form-verify-and-request-load.js
```diff
@@ -0,0 +1,27 @@
+import { statusCodes } from '~/src/server/common/constants/status-codes.js'
+import { findFormBySlug } from '~/src/server/common/forms/services/find-form-by-slug.js'
+
+export async function validateRequestAndFindForm(request, h) {
+  const { slug } = request.params
+
+  if (!slug) {
```
swdpcomputing, line 7: h.response(...).takeover() might be better.

swdpcomputing: I think it would mitigate the config-confirmation bug.

src/server/dev-tools/handlers/clear-application-state.handler.js
aaroncarroll82: This was just to get round the TS lint. It can't ever be not set, but I couldn't convince the linting otherwise. Open to suggestions on how to get round that?

src/server/confirmation/config-confirmation.js
aaroncarroll82: Whoops, good spot. I fixed this on the other controller and hadn't done the same for this one. Will update to the correct property name

src/server/common/helpers/form-verify-and-request-load.js
aaroncarroll82: I can add takeover

src/server/confirmation/config-confirmation.test.js
aaroncarroll82: Yep absolutely right, got to the point where i couldn't see the wood for the trees. Have updated this to check what I actually wanted to check - that when the pre-handler returns an error, the main handler reflects that.

swdpcomputing (approved):

PR #848, opened 2026-04-29
Tag upstream API errors at error level with service/upstreamStatus/at…

swdpcomputing:
When grants-ui-backend goes down (like the failed deployment in dev that returned 502s), grants-ui in front of it returns a 500. The logs say "the BE is the cause" and are emitted at debug level and don't include the upstream HTTP status or which service had failed, so they were filtered out by default and nobody knew why grants-ui was 500ing.                                                                                                                               
                                                        
This PR:

1. Tags every outgoing-call error with structured felds. Every call to grants-ui-backend, GAS, agreements, or consolidated-view now logs service, upstreamStatus, and (on retry-exhaustion) attempts.

2. Promotes those logs from debug to error level. The previous debug() wrapper forced everything down to debug regardless of declared level. This swaps to log() which respects the declared level.                                                                                              
   
So now an alert query like level=error AND service=grants-ui-backend AND upstreamStatus>=500 now matches the next time the BE fails, so we hear about it. User-facing behaviour is unchanged, the 500 page still renders. The alerting-channel investigation (second half of the ticket) is a separate config task.


src/server/agreements/controller.js
```diff
@@ -108,7 +121,7 @@ export const getAgreementController = {
 
       return apiResponse
     } catch (error) {
-      debug(LogCodes.SYSTEM.EXTERNAL_API_ERROR, { endpoint: 'agreements', errorMessage: error.message }, request)
+      logAgreementsUpstreamError(request, error)
```
DavidBarker-EE, line 124: I'm not sure about this one 😂 clearly trying to get around the ESLint rule by encapsulating the log in another function. Let's discuss this with @alanplatt, as it might be wrong to wholly deny `log` in `catch` blocks.

swdpcomputing: Not dodging the rule! log/logger are already on the exclude list in eslint.config.js. Pulled it into a helper because the payload got chunky. Same shape as handleVerificationError in verify-token.js. Happy to inline it, or chat with Alan.

DavidBarker-EE: hehe the function `logAgreementsUpstreamError` is just a wrapper for `log` though right? Reason I mentioned Alan here is that he didn't want catch blocks to be using `log`, maybe that's not the right approach?

alanplatt: We introduced the pattern to tame the errors and make sure that we were not logging multiple times and that debug was actually debug and very obvious. The solution doesnt provided for this use case though unfortunately so i think Steve is on the right lines. I think maybe it should be simpler though with a more and move logUpstreamError to something like helpers/logging/upstream-error.js then add that function to the eslint to keep things in the original spirit



thoughts?

src/server/land-grants/services/land-grants.client.js
```diff
@@ -37,10 +62,15 @@ export async function postToLandGrantsApi(endpoint, body, baseUrl) {
     return response.json()
   }
 
-  return retry(apiOperation, {
+  const result = await retry(apiOperation, {
     timeout: 30000,
     serviceName: `LandGrantsApi.postTo ${endpoint}`
+  }).catch((error) => {
+    logUpstreamError(endpoint, error.status ?? error.code ?? lastUpstreamStatus, error.message, attempts)
```
alanplatt: will this log for every retry?

src/server/agreements/controller.js
swdpcomputing: Agreed.

DavidBarker-EE (approved):

PR #850, opened 2026-04-30
Added a `-v` flag to docker compose down commands to ensure that volu…

DavidBarker-EE:
…mes do not persist between acceptance test runs

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "dev": "run-p frontend:watch server:watch",
     "dev:debug": "run-p frontend:watch server:debug",
     "docker:up": "docker compose up -d",
-    "docker:down": "docker compose down --remove-orphans",
+    "docker:down": "docker compose down --remove-orphans -v",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "dev": "run-p frontend:watch server:watch",
     "dev:debug": "run-p frontend:watch server:debug",
     "docker:up": "docker compose up -d",
-    "docker:down": "docker compose down --remove-orphans",
+    "docker:down": "docker compose down --remove-orphans -v",
```
davidatdefra: For CI using `-v` is fine (although I assumed CI would start with a clean slate each time anyway?), but not sure we always want to remove volumes for dev work. i.e. we'd lose mongo state, etc

There is already a `npm run docker:reset` script to remove all locally built images and volumes if needed.

DavidBarker-EE: Yes, you're right I was a bit over-zealous with that one😂

davidatdefra (approved):

PR #851, opened 2026-05-01
fix(TGC-1278): Serve local form defs via config broker & api

aaroncarroll82:
When running locally via `npm run docker:up` and when running acceptance and journey test suites, serve up the local form definitions via config broker and config API. Note that currently in acceptance tests FORMS_API_SLUGS is still blanked by default, so grants-ui will read direct from file (despite them being available via the broker).

An optional override can be set for the acceptance tests to read from remote config directory, to pull in the latest released version of the config - might be useful to check the tests alongside the latest 'real' version of config at some point


aaroncarroll82:
When running acceptance tests you'll see this, as pulling in local form defs
Screenshot 2026-05-01 at 09 21 27

A local config directory will be created, with the files to be served up by config broker. (git will ignore this directory)
Screenshot 2026-05-01 at 09 21 45

davidatdefra (approved):

PR #858, opened 2026-05-06
TGC-1251 - surface land grants api errors into grants-ui

dcapilla:


src/server/land-grants/services/land-grants.client.js
```diff
@@ -24,11 +24,17 @@ export async function postToLandGrantsApi(endpoint, body, baseUrl) {
     })
 
     if (!response.ok) {
-      await response.arrayBuffer()
+      let message = response.statusText
+      try {
+        const responseBody = await response.json()
+        message = responseBody?.message ?? message
+      } catch {
```
swdpcomputing, line 31: I think this is broken. response.json() already drained the body, so arrayBuffer() will report with TypeError: Body is unusable. that then never hit the throw error below, so callers get a body-consumed error instead of the nice {code, status, message} one.

src/server/land-grants/services/land-grants.client.test.js
```diff
@@ -115,7 +115,7 @@ describe('Land Grants client', () => {
         ok: false,
         status: 500,
         statusText: 'Internal Server Error',
-        arrayBuffer: vi.fn().mockResolvedValue(undefined)
```
swdpcomputing, line 118: I think this mock is incorrect. arrayBuffer resolves cleanly so the bug above passes. Change it to arrayBuffer: vi.fn().mockRejectedValue(new TypeError('Body is unusable')) and it'll fail until the client gets fixed.

src/server/land-grants/services/land-grants.client.js
dcapilla: ah good catch! I think you are right

swdpcomputing (approved):

PR #861, opened 2026-05-07
TGC-1272: Task list update

davidatdefra:


src/server/common/controllers/question-page-with-parcel-check.controller.js
```diff
@@ -1,12 +1,12 @@
...
-import { log, debug, LogCodes } from '~/src/server/common/helpers/logging/log.js'
+import { debug, log, LogCodes } from '~/src/server/common/helpers/logging/log.js'
 import { fetchParcelsFromDal } from '~/src/server/common/services/consolidated-view/consolidated-view.service.js'
 import { getCachedAuthParcels, setCachedAuthParcels } from '~/src/server/land-grants/services/parcel-cache.js'
 import { stringifyParcel } from '~/src/server/land-grants/utils/format-parcel.js'
 import { SystemError } from '~/src/server/common/utils/errors/SystemError.js'
+import TaskPageController from '~/src/server/task-list/task-page.controller.js'
 
-export default class QuestionPageWithParcelCheckController extends QuestionPageController {
+export default class QuestionPageWithParcelCheckController extends TaskPageController {
```
dcapilla: I just find it a bit confusing that this QuestionPageWithParcelCheckController is now a TaskPageController since this controller might be applied to individual question pages that are not related to any tasks. I know this will work, just the naming confuses me a bit.

dcapilla (dismissed):

src/server/common/forms/definitions/example-grant-with-task-list-hide-questions.yaml
```diff
@@ -0,0 +1,354 @@
...
+          <p class="govuk-body">This is some example help text.</p>
+          <p class="govuk-body">It can include links to guidance or other information.</p>
+        options:
+          position: below # Renders component below task list
+
+  # Task pages - Example task one
+  - title: Example multiple components
+    id: 523058ea-46f0-44fd-ae79-1a6be652c7f9
+    path: /multiple-components-task-page
+    section: 52757632-60cb-437f-a648-cfbc6acdacd0
```
dcapilla, line 182: wouldn't be more human friendly if we linked sections by `name` (slug)? so this would be 

```
section: example-task-one
```

src/server/task-list/views/task-list-page.html
```diff
@@ -16,19 +16,19 @@ <h1 class="govuk-heading-l">{{ pageTitle }}</h1>
     {% if showCompletionStatus !== false %}
     <h2 class="govuk-heading-s govuk-!-margin-bottom-2">Application status</h2>
     <p class="govuk-body govuk-!-margin-bottom-7">
-      You have completed {{ completionStats.completed }} of {{ completionStats.total }} tasks.
+      You have completed {{ completionStats.completed }} of {{ completionStats.total }} task pages.
```
dcapilla, line 19: this will also display if showQuestions = false but the number no longer matches the list of tasks displayed


Image


This is not broken, but might confuse users. Do we really want this or do we want to display the number of tasks rather than the number of pages?

src/server/task-list/task-list.helper.js
```diff
@@ -270,75 +270,196 @@ export function getTaskListPath(model) {
...
+    const href = hrefPath ? `${basePath}${hrefPath}` : undefined
+
+    const taskItem = createTaskItemBase(taskTitle)
+
+    if (status === 'completed') {
+      taskItem.href = href
+      taskItem.status = createStatusTag(statuses.completed, 'Completed', 'govuk-tag--green')
+    } else if (status === 'inProgress') {
+      taskItem.href = href
+      taskItem.status = createStatusTag(statuses.inProgress, 'In progress', 'govuk-tag--blue')
```
dcapilla: be good to extract these strings to an object so it is easy to find the task statuses and can easily be imported if needed.  Even the classes if you want, just an idea anyway:

```
const TAG_CLASSES = {
  green: 'govuk-tag--green',
  blue: 'govuk-tag--blue',
  yellow: 'govuk-tag--yellow',
  grey: 'govuk-tag--grey',
  purple: 'govuk-tag--purple'
}

const TASK_STATUS = {
  completed: { text: 'Completed', classes: TAG_CLASSES.green },
  inProgress: { text: 'In progress', classes: TAG_CLASSES.blue },
  notStarted: { text: 'Not started', classes: TAG_CLASSES.yellow },
  cannotStart: { text: 'Cannot start yet', classes: TAG_CLASSES.grey },
  cannotContinue: { text: 'On hold', classes: TAG_CLASSES.purple }
}
```

src/server/task-list/task-list.helper.js
dcapilla: I know is not part of this change so feel free to ignore.

This function and mapComponentToViewModelComponent are view model concerns rather than a utility function. Would be better placed on the controller or a separate tasklist.view-model.js file

src/server/common/forms/definitions/example-grant-with-task-list-hide-questions.yaml
davidatdefra: DXT rules mean they have to be UUIDs 😢

src/server/common/controllers/question-page-with-parcel-check.controller.js
davidatdefra: Yeah, it is a little odd. `TaskPageController` handles not being in a task list fine. Maybe could rename to... something? `TaskEnabledQuestionPageController` ?

src/server/task-list/views/task-list-page.html
davidatdefra: Yes. I did change the text a little, so it now says "X of Y task **pages**" rather than just tasks, but maybe could be improved.

I actually think it's more useful in this mode as you can only see "In progress" in the task list, so you've no idea how many questions/pages you have left to answer.

src/server/common/controllers/question-page-with-parcel-check.controller.js
dcapilla: Yes, I think this is more a renaming thing than anything else. Something that suggests this is a Question page would work perhaps, like the one you suggested! Thanks :)

src/server/task-list/views/task-list-page.html
dcapilla: Yes I see your point, perhaps would be good if we get some UX review on this but doesn't need to happen now as part of this PR!

src/server/common/controllers/question-page-with-parcel-check.controller.js
davidatdefra: I was using Claude to come up with a better name, and I explained to it the purpose of the controller and it suggested refactoring to use a `mixin` pattern. I looked into it and thought it would make sense, so implemented it

dcapilla (approved):

PR #868, opened 2026-05-11
TGC-1289: Check details terminal routing

davidatdefra:


DavidBarker-EE:
Looks good, I'm not sure if you want to ignore or fix the sonar errors, so I'll leave it un-approved for the moment

src/server/status/status-helper.js
```diff
@@ -332,8 +333,17 @@ export const formsStatusCallback = async (request, h, context) => {
   const grantRedirectRules = request.app.model?.def?.metadata?.grantRedirectRules
   const isWithinGrantPages = request.headers['sec-fetch-site'] === 'same-origin'
 
+  const checkDetailsChangesPending = context.state.checkDetailsChangesPending
```
swdpcomputing, line 336: Are we sure context.state is always defined here? Everywhere else seems to assume it is, but if it's ever undefined this throws before we even get to the status logic. Worth a ?. just in case?

davidatdefra: `context.state` will always be at least `{}` by this point (our function called during `onRequest`) even if it would be null, as forms-engine-plugin ensures it is a blank object and more likely will also contain at least the generated `$$__referenceNumber` as that is also injected by DXT if none exists

swdpcomputing (approved):

PR #885, opened 2026-05-18
remove gas config from grants-ui

swdpcomputing:


davidatdefra (commented):
Need to also remove reference to schema from `config.http`

tools/compare-remote-config.sh
```diff
@@ -3,10 +3,18 @@ LOCAL_FILE="src/server/common/forms/definitions/example-grant-with-auth.yaml"
 
 echo "example-grant-with-auth config has changed - checking grant-config-example-grants repo to see if local config change is reflected there..."
 REMOTE_CONFIG_UPDATED_TO_MATCH=false
+# Normalise the local file through yq once so remote comparisons (also yq-normalised
+# below) aren't tripped up by hand-formatted whitespace or quoting differences.
+yq '.' "$LOCAL_FILE" > local.normalized.yaml
 # get latest tagged file, and compare
 TAG=$(curl -s https://api.github.com/repos/DEFRA/grant-config-example-grants/tags | jq -r '.[0].name')
```
DavidBarker-EE: If this script is moved into a new repo, this can silently fail (empty list, gh rate limit or error). Adding the `--fail` flag to curl would sort the errors and checking for empty tags might be a good idea.

```
if [ -z "$TAG" ] || [ "$TAG" = "null" ]; then
  echo "❌ Could not determine latest tag for DEFRA/grant-config-example-grants"
  exit 1
fi
```

tools/compare-remote-config.sh
```diff
@@ -3,10 +3,18 @@ LOCAL_FILE="src/server/common/forms/definitions/example-grant-with-auth.yaml"
 
 echo "example-grant-with-auth config has changed - checking grant-config-example-grants repo to see if local config change is reflected there..."
 REMOTE_CONFIG_UPDATED_TO_MATCH=false
+# Normalise the local file through yq once so remote comparisons (also yq-normalised
+# below) aren't tripped up by hand-formatted whitespace or quoting differences.
+yq '.' "$LOCAL_FILE" > local.normalized.yaml
```
DavidBarker-EE: If our runner was short-lived, this probably wouldn't cause potential problems, but we're writing files to the repo root here, and our runner is long-lived. Maybe we should create a temporary directory to house these files and then clean them up before exiting.

tools/compare-remote-config.sh
davidatdefra: Might also be worth check `yq` is actually installed first and error with message if not - in the default GH runner it currently is, but only because it comes with the latest Ubuntu image, which might change.

tools/compare-remote-config.sh
```diff
@@ -1,12 +1,35 @@
...
 # get latest tagged file, and compare
-TAG=$(curl -s https://api.github.com/repos/DEFRA/grant-config-example-grants/tags | jq -r '.[0].name')
-curl -sL https://raw.githubusercontent.com/DEFRA/grant-config-example-grants/$TAG/example-grant-with-auth/grants-ui/example-grant-with-auth.yaml -o tagged.yaml
-if ! diff -q "$LOCAL_FILE" tagged.yaml > /dev/null; then
+TAG=$(curl -sS --fail https://api.github.com/repos/DEFRA/grant-config-example-grants/tags | jq -r '.[0].name')
+if [ -z "$TAG" ] || [ "$TAG" = "null" ]; then
+  echo "❌ Could not determine latest tag for DEFRA/grant-config-example-grants"
+  exit 1
+fi
+curl -sL https://raw.githubusercontent.com/DEFRA/grant-config-example-grants/$TAG/example-grant-with-auth/grants-ui/example-grant-with-auth.yaml -o "$TMPDIR/tagged.yaml"
```
DavidBarker-EE: minor pedantry,  but a `--fail` on this would be good too

DavidBarker-EE (dismissed):

DavidBarker-EE (approved):

PR #887, opened 2026-05-18
fix decimal point error in total-area-of-woodland

swdpcomputing:


src/server/woodland/woodland-hectares-page.controller.js
```diff
@@ -15,6 +15,8 @@ const HECTARES_UNDER_TEN_FIELD_NAME = 'hectaresUnderTenYearsOld'
 
 const ERROR_BELOW_MINIMUM = `The total area of woodland must be at least ${MIN_WOODLAND_TOTAL_AREA_HA}ha`
 
+const roundTo4dp = (/** @type {number} */ value) => Math.round(value * 10000) / 10000
```
dcapilla: could we use Math.trunc here instead please? GAS will check that totalHectaresForSelectedParcels matches the individual areas of the land parcels we submit. So if we round this there might be a risk that they don't match. This is probably not the case because I don't think you are saving this into the state, but still.

dcapilla (approved):

PR #888, opened 2026-05-19
Enforce submit permissions for Countryside Stewardship applications

Andrew-Folga:
Implements submit permission checks for Countryside Stewardship applications.

Users without the required DAL submit permission are now prevented from submitting applications from the declaration page and are redirected to a dedicated "cannot submit" page.

## Changes

- added generic `requirePermission` permission guard helper
- added `requireSubmitCsApplication` submit permission guard
- enforced submit permission checks in `DeclarationPageController`
- added `/cannot-submit` route and view
- added return navigation support back to the application/task list
- wired permission handling into CS routing

Users with submit permission:
- can successfully submit applications

Users without submit permission:
- cannot submit applications
- are redirected to the cannot-submit page
- can return to the application/task list

### DIFF
--- a/src/server/common/helpers/permissions/permission-rules.js
+++ b/src/server/common/helpers/permissions/permission-rules.js
@@ -7,34 +7,5 @@ export const permissionRules = {
       amend: 'AMEND',
       submit: 'SUBMIT'
     }
-  },
-
-  csAgreements: {
-    permissionGroup: 'COUNTRYSIDE_STEWARDSHIP_AGREEMENTS',
-
-    permissions: {
-      view: 'VIEW',
-      amend: 'AMEND',
-      submit: 'SUBMIT'
-    }
-  },
-
-  landDetails: {
-    permissionGroup: 'LAND_DETAILS',
-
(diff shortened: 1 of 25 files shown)


src/server/common/helpers/permissions/permission-rules.js
dcapilla, line 1: shouldn't this come from config instead of having this? I know we haven't integrated config broker yet, but might be easier if we put this as part of the grant definition or something that is easier to migrate later

Andrew-Folga: I am not sure I fully understand - at the moment this config would be the same regardless of the grant definition (Woodland, SFI/Farm Payments, etc.) since they all currently use the same permission groups/actions.

dcapilla (approved):

src/server/common/helpers/permissions/permission-rules.js
dcapilla: right ok, for some reason I thought permissions would be configurable per journey but it might not be the case. Food for thought!

PR #892, opened 2026-05-20
TGC-1310: Modularise mongo

davidatdefra:


dcapilla (approved):

davidatdefra:
Not required - agreed to use `MONGO_URI` for both connections as simplifies CDP config

PR #894, opened 2026-05-21
new chunk of type improvements up to src/server/common/helpers/whitelist/whitelist.js

swdpcomputing:


src/server/common/helpers/print-application-service/utils/build-print-payment-view-model.js
```diff
@@ -63,3 +63,32 @@ export function buildPrintAdditionalPayments(payment) {
     }
   ])
 }
+
```
dcapilla: I am fine with this but you might want to move these to a `.d.js` file like we do for other types. Feel free to resolve this comment and merge without changes.

dcapilla (dismissed):

src/server/common/helpers/print-application-service/utils/build-print-payment-view-model.js
swdpcomputing: Abstracted.

dcapilla (approved):

PR #901, opened 2026-05-26
TGC-1360: Update details not found

davidatdefra:


src/server/details-page/check-details.controller.js
```diff
@@ -172,6 +181,8 @@ export default class CheckDetailsController extends QuestionPageController {
       )
```
alanplatt, line 187: could check model.pages is not unexpectedly empty so if DXT changes the forms engine to async we catch it? not too fussed about that so up to you

src/server/details-page/check-details.controller.js
```diff
@@ -172,6 +181,8 @@ export default class CheckDetailsController extends QuestionPageController {
...
       model.pages.splice(insertAt, 0, updateDetailsController)
       model.pageMap?.set(UPDATE_DETAILS_PATH, updateDetailsController)
+
+      log(LogCodes.SYSTEM.CHECK_DETAILS_TERMINAL_PAGE_INJECTED, {})
```
alanplatt: include which grant it was called from?

alanplatt (approved):

PR #911, opened 2026-05-29
Feature/tgc 1373 centralise permission checks

Andrew-Folga:
This PR introduces a centralised permission enforcement pipeline for grant applications.

Previously, permission checks were distributed across routes, page definitions, and controller behaviour. This change consolidates authorisation logic into a single request pipeline step responsible for:

- resolving required page permissions
- enforcing access rules consistently
- handling amend-only users
- restricting view-only users
- supporting permission enforcement toggles via config
- preparing the codebase for future multi-resource permission support

The implementation now provides a single source of truth (grant's yaml config under `permissions` that now includes `pageAccess` section) for application access behaviour.

This PR enables user permissions for `example-grant-with-auth` only.


src/plugins/permissions.js
```diff
@@ -3,7 +3,7 @@ import { can } from '~/src/server/common/helpers/permissions/can.js'
 const EXCLUDED_PATHS = ['/health', '/assets', '/public', '/auth']
 
 function shouldSkip(request) {
-  return EXCLUDED_PATHS.some((path) => request.path.startsWith(path))
+  return EXCLUDED_PATHS.some((path) => request.path.includes(path))
 }
```
swdpcomputing, line 7: This is a security hole. Any grant page whose URL happens to contain the letters public, auth, assets, or health would silently bypass the entire permission system. Unlikely but still a hole.

swdpcomputing:
Is permission-logger.test.js still needed?

src/server/common/helpers/permissions/page-permissions.js
```diff
@@ -0,0 +1,17 @@
...
+
+  const matchedRule = permissionConfig?.pageAccess?.rules?.find((rule) => rule.paths.includes(request.params.path))
+
+  return matchedRule?.permission ?? permissionConfig?.pageAccess?.default
+}
+
+export function getPermissionResource(request) {
+  const permissionConfig = getPermissionConfig(request)
+
+  return permissionConfig.resource || 'csApplications'
```
swdpcomputing: If a future grant turns on enforcement but forgets to set its resource, would it it silently inherit Countryside Stewardship's permission rules instead of failing loudly?

src/server/common/request-pipeline/permissions/enforce-page-permission.js
```diff
@@ -0,0 +1,153 @@
...
+    const model = request.app.model
+    if (!model) {
+      throw forbidden('Form model missing')
+    }
+    const returnUrl = getReturnToApplicationPath(model, basePath)
+    const redirectUrl = returnUrl ? `/cannot-submit?returnUrl=${encodeURIComponent(returnUrl)}` : '/cannot-submit'
+    return h.redirect(redirectUrl).takeover()
+  }
+
+  throw forbidden('Insufficient permissions')
```
swdpcomputing, line 194: Did we mean to drop the permission logging here? The old guards used to fire logPermissionEvent on success/failure/bypass, but this gate just throws forbidden silently . permission-logger.js and the PERMISSIONS log codes look orphaned now. Was that intentional, or should the deny path still log? Feels like something we'd want to keep.

src/plugins/permissions.js
Andrew-Folga: Yes it was initially startsWith and reverting it to that now.

Andrew-Folga: also removing assets and public as they are already unauthenticated so would be checked earlier in `!request.auth.isAuthenticated`

src/server/common/helpers/permissions/page-permissions.js
Andrew-Folga: Good spot and fixed now.

src/plugins/permissions.js
Andrew-Folga: also improved the testing of the EXCLUDED_PATHS to handle `/` in `auth/` to make sure we exclude exact paths only

src/server/common/request-pipeline/permissions/enforce-page-permission.js
Andrew-Folga: this has been pushed this morning  already

Andrew-Folga:
@swdpcomputing Thanks for your review - all comments addressed. In terms of your question re: Is permission-logger.test.js still needed? - yes it is needed as it has some logic inside.

swdpcomputing (approved):

PR #912, opened 2026-05-29
added audit event to grant start

swdpcomputing:


compose.yml
```diff
@@ -78,6 +78,13 @@ services:
       FORMS_API_CACHE_TTL_SECONDS: ${FORMS_API_CACHE_TTL_SECONDS:-300}
       SFD_UPDATE_URL: '${SFD_UPDATE_URL:-http://localhost:3000/sfd/mock}'
       SFD_UPDATE_ENABLED: ${SFD_UPDATE_ENABLED:-false}
+      AWS_REGION: ${AWS_REGION:-eu-west-2}
```
davidatdefra: You don't need to put these env vars here:
```
AWS_REGION
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
```
Instead, just above the `environment:` key put:
```
    env_file:
      - 'localstack/aws.env'
```
That contains those values and is used across multiple containers in the compose.

swdpcomputing: Edited.

src/server/common/helpers/audit/audit.js
```diff
@@ -0,0 +1,83 @@
...
+        environment: mapEnvironment(config.get('cdpEnvironment')),
+        generateCorrelationId: true
+      }
+
+      server.ext('onPreResponse', (request, h) => {
+        if (isSuccessfulGrantAccess(request)) {
+          const slug = request.params.slug
+          const event = buildAuditEventForGrantAccess(request)
+
+          publishAuditEvent(event, publisherConfig)
```
dcapilla, line 63: do we need to publish an event every time `isSuccessfulGrantAccess` is true or is this a one-off operation per journey per user?

dcapilla (dismissed):

src/server/common/helpers/audit/audit.js
swdpcomputing: Every time, I imagine.

dcapilla (approved):

PR #917, opened 2026-06-03
TGC-1381 - Ability to clear state from /agreement page + pass grantCode/clientRef to agreements proxy

dcapilla:
Clears the grants-ui-backend state document when a user navigates to the agreements proxy page and triggers a state reset (via the "Clear application state" link).

**Changes**

- Adds clearSavedStateFromApi call in clearApplicationStateHandler for the no-slug path (agreements context), using sbi:grantCode key from yar and a minted lock token
- Stores { grantCode, clientRef } in yar under YarKeys.GRANT_APPLICATION_CONTEXT after post-submission redirect, so the agreements page and dev-tools handler have the context needed to clear state
- Passes new context variables grantCode and clientRef to the agreements proxy through request headers
- Extracts clearStateWithSlug and clearStateWithoutSlug helpers to reduce function complexity


src/server/dev-tools/handlers/clear-application-state.handler.js
```diff
@@ -19,37 +23,82 @@ import { clearParcelCache } from '~/src/server/land-grants/services/parcel-cache
...
+  const { grantCode } = /** @type {{ grantCode: string | undefined }} */ (
+    request.yar?.get(YarKeys.GRANT_APPLICATION_CONTEXT) || {}
+  )
+
+  if (!sbi || !grantCode || !contactId) {
+    return
+  }
+
+  const grantVersion =
+    /** @type {{ metadata?: { version?: string | number } }} */ (request.app.model?.def)?.metadata?.version ?? 1
```
alanplatt: You're going to need the version of the grant here as we use grantVersion in the mongo lookup. If there is already a lock etc with a version of 1.0.0 then defaulting to 1 will break

alanplatt: also we're moving away from the int 1 to semver

src/server/common/request-pipeline/redirects/forms-status-redirect.js
```diff
@@ -264,7 +265,10 @@ async function handlePostSubmission(request, h, context, previousStatus, grantCo
 
   await persistStatus(request, rule.toGrantsStatus, previousStatus, grantId, context.state)
 
-  const redirectUrl = rule.toPath === agreements.get('baseUrl') ? rule.toPath : buildRedirectUrl(grantId, rule.toPath)
+  const isAgreementsRedirect = rule.toPath === agreements.get('baseUrl')
+  const redirectUrl = isAgreementsRedirect ? rule.toPath : buildRedirectUrl(grantId, rule.toPath)
+
+  request.yar.set(YarKeys.GRANT_APPLICATION_CONTEXT, { grantCode, clientRef: clientRef.toLowerCase() })
```
alanplatt: request.yar.set we should make sure we clear this so that we don't end up with details that could potentially pollute another application

src/server/dev-tools/handlers/clear-application-state.handler.js
```diff
@@ -19,37 +23,82 @@ import { clearParcelCache } from '~/src/server/land-grants/services/parcel-cache
...
+  } catch (error) {
+    const sessionKey = cacheService._Key(request)
+    const sessionError = new SessionError({
+      message: 'Session state clear failed',
+      source: 'clearApplicationStateHandler',
+      reason: 'session_state_clear_failure',
+      slug,
+      sessionKey
+    })
+    throw sessionError.from(/** @type {Error} */ (error))
```
alanplatt: This should follow the pattern like here https://github.com/DEFRA/grants-ui/pull/917/changes#diff-4c048c660423d6b521b168608ca3687c19786ed2f6aa1a6ca2b3ea1eae73ef9aR98

alanplatt (approved):

PR #931, opened 2026-06-09
View users should be redirected to confirmation once application is submitted and they access slug via SFD tile

Andrew-Folga:


src/server/common/request-pipeline/permissions/enforce-page-permission.js
```diff
@@ -120,57 +119,53 @@ export function enforcePagePermission(request, h, context) {
...
 
-  logPermissionEvent({
-    request,
-    grantCode,
-    permission: requiredPermission,
-    enforcementEnabled: true,
-    authorised: false
-  })
+  // 4. view-only users post-submission -> go to confirmation
+  if (isViewOnlyUser(request, resource) && isSubmittedApplication(context)) {
```
dcapilla, line 162: what happens if application is submitted and agreement has been drafted? In that case the expectation is to be redirected to /agreement so I am not sure this code would work in that scenario.

Wouldn't it be better to delegate the redirect logic to whoever handles it already and apply permissions after the user has been redirected?

PR #932, opened 2026-06-09
TGC-1399: Update grants-ui compose - land grants now uses SQS

davidatdefra:
Also includes updates for config broker enabled grants-ui-backend


localstack/start-localstack.sh
alanplatt, line 1: do you think we should use set -e so that the scripts stops on errors? they'll be hidden otherwise

alanplatt:
do we need to add depends_on: localstack to grants-ui-backend now?

swdpcomputing (approved):

PR #947, opened 2026-06-13
TGC-1410: Enforce single active session per user

alanplatt:
Adds a second Catbox cache partition (user-session-index) that maps contactId → sessionId. On sign-in, any existing session for the same user is dropped from the main cache before the new session is stored, preventing concurrent sessions. The index entry is cleared on sign-out.

Works transparently across HA instances via shared Redis, and does not force re-authentication of currently signed-in users at deploy time.

### DIFF
--- a/src/server/auth/single-session.js
+++ b/src/server/auth/single-session.js
@@ -1,0 +1,35 @@
+import { log, LogCodes } from '~/src/server/common/helpers/logging/log.js'
+
+/**
+ * Ensures only one active session exists per user. If an existing session is found
+ * in the index for this contactId, it is dropped from the main cache before the new
+ * session is stored. The index is then updated to point at the new sessionId.
+ *
+ * The guard `existingSessionId !== newSessionId` prevents a re-entrant sign-in from
+ * evicting the very session it is about to create.
+ *
+ * @param {SessionCache} cache - main session cache (server.app.cache)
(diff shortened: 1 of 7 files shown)


src/server/auth/single-session.js
```diff
@@ -0,0 +1,35 @@
...
+ * evicting the very session it is about to create.
+ *
+ * @param {SessionCache} cache - main session cache (server.app.cache)
+ * @param {UserSessionIndex} userSessionIndex - contactId→sessionId index (server.app.userSessionIndex)
+ * @param {string} contactId - stable user identifier from the JWT payload
+ * @param {string} newSessionId - UUID generated for the incoming sign-in
+ * @returns {Promise<void>}
+ */
+export async function invalidatePreviousSession(cache, userSessionIndex, contactId, newSessionId) {
+  const existingSessionId = /** @type {string | null} */ (await userSessionIndex.get(contactId))
```
dcapilla, line 18: this get, drop, set sequence is not atomic. So concurrent sign ins from the same user can end setting two sessionIds (meaning one session leaks past its invalidation) for the same contactId.

Might be worth adding some lock mechanism to prevent this. Not a blocker anyway, just a suggestion.

alanplatt:
This was originally to cover an issue raised during a securtity audit. It turns out its an accepted thing across the Defra estate to allow simultaneous logins so I'm closing this PR.

PR #961, opened 2026-06-16
Add support for semver to GAS submissions

DavidBarker-EE:


src/server/common/helpers/grant-application-service/state-to-gas-payload-mapper.js
```diff
@@ -13,6 +14,28 @@
...
+  }
+}
+
+/**
+ * Resolves the grant configuration version to send to GAS.
+ * @param {object} request - Hapi request
+ * @returns {string}
+ */
+export const resolveGasConfigVersion = (request) => {
+  const configVersion = request?.app?.model?.def?.metadata?.version ?? 1
```
davidatdefra: If we're using a "legacy" repo based form def, does that fallback of `1` mean it will fail the semver check and not submit?

I can't parse the regex in my head so maybe that is covered?

The backend is happy to accept `1` and transform it to `1.0.0` to handle legacy form defs, but as we're sending to GAS directly after this, maybe just make the fallback `1.0.0` in this case if it would otherwise fail when it's `1`

swdpcomputing: should this default be the string '1.0.0' rather than the number 1? right now ?? 1 hands a number straight into assertSemverConfigVersion, which only accepts strings, so it throws for any grant without a version.

.gitignore
```diff
@@ -24,3 +26,4 @@ localstack/config-broker-local
 /.grants-ui-cli-state.json
 /acceptance/schemas/
 /woodland-grant-journey-tests-schemas/
+/docs
```
swdpcomputing: any reason this is in here? that folder's already tracked so this'd quietly start ignoring new docs additions

src/server/common/helpers/grant-application-service/state-to-gas-payload-mapper.js
DavidBarker-EE: Good spot, I actually had code here to check for that but deemed it too defensive. I need to modify this. The regex doesn't account for 1.

davidatdefra (approved):

PR #974, opened 2026-06-19
Use ConfirmationPageController.

Andrew-Folga:
Use ConfirmationPageController that taps into the dxt lifecycle instead of grants using the config confirmation hapi route handler that is missing the internals of the onRequest callback. The onRequest callback that runs forms request pipeline runs redirections and also has user permissions checks centralised which we wanted to achieve across the repo.

This also removes the old hapi route config-confirmation.js handler and its corresponding template - all is migrated to ConfirmationPageController.


src/server/confirmation/confirmation-page.controller.js
```diff
@@ -14,6 +18,45 @@ export default class ConfirmationPageController extends StatusPageController {
...
   }
 
+  /**
+   * Loads and validates confirmation content for the form
+   * @param {AnyFormRequest} request - The request object
+   * @param {object} state - the DXT state object containing application details
+   * @returns {Promise<object|null>} Content result with confirmationContent and formDefinition
+   */
+  async loadConfirmationContent(request, state) {
+    const form = this.model.def
```
swdpcomputing, line 28: Is this going to trip the service guard in src/server/confirmation/services/confirmation.service?

Does the custom confirmation text now shows up blank on every form (the content loader gets handed the wrong object and quietly gives up)?

Andrew-Folga: Thanks.

Can you please explain to me why do we need this guard at all @swdpcomputing ?

Andrew-Folga: I think the need for the guard was removed with this change: [https://github.com/DEFRA/grants-ui/commit/a317eba8cf00e1274f64df15e700f231ed005db4#diff-dc8ee216393a789632dcd64b22[…]805604795bee4d0b28a47aaeec3e482](https://github.com/DEFRA/grants-ui/commit/a317eba8cf00e1274f64df15e700f231ed005db4#diff-dc8ee216393a789632dcd64b2219c921d805604795bee4d0b28a47aaeec3e482)

Andrew-Folga: This has been discussed and addressed by removing the `form.id` guard as the function does not use the id anymore.

dcapilla (approved):

PR #987, opened 2026-06-23
TGC-1294 - New allowlist endpoint

dcapilla:
Adds a new /allowlist/grants endpoint integration for per-grant access control, controlled by BACKEND_ALLOWLIST_ENABLED_SLUGS. Grant codes listed in that flag are checked against the grants-ui-backend allowlist endpoint instead of the existing env-var whitelists. Any grant code not in the list falls back to the previous env var whitelist behaviour unchanged, making this a safe incremental rollout with no impact on existing grants.


src/server/common/helpers/allowlist/allowlist.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  if (!request.params.slug) {
+    return h.continue
+  }
+
+  const crn = /** @type {string} */ (request.auth.credentials.crn)
+  const sbi = /** @type {string} */ (request.auth.credentials.sbi)
+
+  const allForms = await getAllForms()
```
davidatdefra: Can simplify from here to Line 46 - we decided to remove `grantCode` from the YAML as `filename` is the "source of truth" for both `grantCode` and `slug`, so if you've got `request.params.slug`, that *is* the `grantCode`. So... we don't really need to lookup allForms and metadata.

src/server/common/forms/services/form.js
```diff
@@ -240,6 +240,16 @@ export function validateDetailsPageConfiguration(form, definition) {
  * @returns {void}
  */
 export function validateWhitelistConfiguration(form, definition) {
+  const enabledCodes = /** @type {string[]} */ (config.get('forms.backendAllowlistEnabledSlugs'))
+
+  const grantCode =
```
davidatdefra: Same as previous comment - we can always assume slug === grantCode

src/server/common/helpers/whitelist/whitelist.js
```diff
@@ -26,7 +27,18 @@ const whitelistHandler = async (request, h) => {
   const sbi = /** @type {string} */ (request.auth.credentials.sbi)
 
   const allForms = await getAllForms()
-  const grantMetadata = allForms.find((form) => form.slug === request.params.slug)?.metadata
+  const form = allForms.find((f) => f.slug === request.params.slug)
```
davidatdefra: Can simplify as in other comments - slug === grantCode

davidatdefra (approved):

PR #995, opened 2026-06-25
add JSDoc types across modules, centralise duplicate typedefs, and re…

swdpcomputing:
…place the semver regex with the semver package

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "nunjucks": "3.2.4",
     "pino": "9.14.0",
     "pino-pretty": "13.1.3",
+    "semver": "7.8.4",
     "undici": "8.5.0",
     "yaml": "2.9.0"
   },
@@ -143,6 +144,7 @@
     "@types/lodash": "4.17.24",
     "@types/node": "25.9.3",
     "@types/nunjucks": "3.2.6",
+    "@types/semver": "7.7.1",
     "@types/webpack-assets-manifest": "5.1.4",
     "@typescript-eslint/eslint-plugin": "8.61.1",
     "@typescript-eslint/parser": "8.61.1",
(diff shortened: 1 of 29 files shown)


package.json
```diff
@@ -143,6 +143,7 @@
     "@types/lodash": "4.17.24",
     "@types/node": "25.9.3",
     "@types/nunjucks": "3.2.6",
+    "@types/semver": "7.7.1",
```
davidatdefra, line 147: This should be a runtime dependency, not a dev dependency

package.json
davidatdefra: I meant this one

src/server/common/helpers/grant-application-service/state-to-gas-payload-mapper.js
```diff
@@ -14,26 +16,27 @@
...
-const SEMVER_PATTERN =
-  /^(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)(?:-(?:(?:0|[1-9]\d*|\d*[A-Za-z-][\dA-Za-z-]*)(?:\.(?:0|[1-9]\d*|\d*[A-Za-z-][\dA-Za-z-]*))*))?(?:\+(?:[\dA-Za-z-]+(?:\.[\dA-Za-z-]+)*))?$/
-
+/**
+ * @param {unknown} configVersion - the grant configuration version to validate
+ * @returns {void}
+ */
 const assertSemverConfigVersion = (configVersion) => {
-  if (typeof configVersion !== 'string' || !SEMVER_PATTERN.test(configVersion)) {
+  if (typeof configVersion !== 'string' || semver.valid(configVersion) === null) {
```
davidatdefra: I think this needs to be stricter - just reading the semver docs, this will not fail `v1.0.0` or pre-release type versions like `1.0.0-beta.1`, but config broker version only supports strict `major.minor.patch`

davidatdefra (dismissed):

davidatdefra (approved):

PR #998, opened 2026-06-26
TGC-1451: Update CLI and compose

davidatdefra:


compose.yml
```diff
@@ -92,22 +92,29 @@ services:
...
     image: defradigital/grants-ui-backend
-    pull_policy: always
+    pull_policy: missing
     depends_on:
       localstack:
         condition: service_healthy
       mongodb:
         condition: service_healthy
       mongo-ready:
-        condition: service_healthy
```
alanplatt, line 111: we should keep this for people not using the cli

compose.yml
```diff
@@ -188,14 +197,15 @@ services:
...
+      interval: 30s
+      start_interval: 1s
+      timeout: 3s
+      retries: 3
+      start_period: 10s
 
   grants-ui-dal-stub:
     image: defradigital/grants-ui-dal-stub:latest
-    pull_policy: always
+    pull_policy: missing
```
alanplatt: this will mean it will silently become out of date for people, is that what we want?

compose.land-grants.yml
```diff
@@ -10,6 +10,9 @@ services:
       - ./localstack/start-localstack-land-grants.sh:/etc/localstack/init/ready.d/start-localstack-land-grants.sh
 
   grants-ui:
+    depends_on:
+      land-grants-backend:
+        condition: service_started
```
alanplatt: may cause issue with automation

compose.yml
```diff
@@ -5,11 +5,11 @@ services:
       target: ${BUILD_TARGET:-development}
     depends_on:
       redis:
-        condition: service_healthy
+        condition: service_started
```
alanplatt: this is going to cause issues with automated tests

alanplatt: same below

tools/grants-ui-cli.js
```diff
@@ -1026,6 +1125,31 @@ async function main() {
...
+          (k) => previousLocalServices.includes(k) !== newLocalServices.includes(k)
+        )
+        const runningSet = new Set(getRunningServices())
+        const servicesToRestart = changedKeys
+          .map((k) => LOCAL_SERVICES.find((s) => s.key === k)?.composeService)
+          .filter((name) => name && runningSet.has(name))
+
+        if (servicesToRestart.length) {
+          pauseStdin()
+          const restartStatus = cmdRestart(servicesToRestart, dryRun)
```
DavidBarker-EE, line 1149: Just want to check this is the desired behaviour. `cmdRestart` will call `docker compose restart --no-deps <...services>`. From what I see here we actually want this to rebuild as well, not just restart, in which case `--force-recreate <...services>` will be needed.

compose.yml
```diff
@@ -92,22 +92,29 @@ services:
...
+      interval: 30s
+      start_interval: 1s
+      timeout: 3s
+      retries: 3
+      start_period: 30s
 
   grants-ui-backend:
     image: defradigital/grants-ui-backend
-    pull_policy: always
+    pull_policy: missing
```
DavidBarker-EE: Switching this to missing will mean existing local build tags will not be refreshed in any way and will just sit and become stale. Is this strictly necessary? Seems to me `always` is the better policy here?

DavidBarker-EE: With this as is, even with `docker compose down --remove-orphans --rmi local`, the `latest` tag will always use the local version and not pull the new one.

compose.yml
davidatdefra: Not sure what this means? Even if not using the CLI, the replicaset script will still mean this service "completes" and never returns a "healthy" status. This is on purpose as it's a "one shot" startup thing, not meant to hang around. It used to force staying running as compose didn't used to support `service_completed_successfully` but has done for a while now.

alanplatt: sorry, read it wrong

tools/grants-ui-cli.js
```diff
@@ -755,8 +809,8 @@ ${BOLD}Commands:${RESET_COLOR}
   up      Start containers
   down    Stop containers (uses saved state — no need to re-select)
   debug   Restart grants-ui in debug mode (detached, port 9229)
+  restart Restart running containers (selectable; uses --no-deps)
```
alanplatt, line 812: should update with --force-recreate

compose.yml
```diff
@@ -287,18 +302,21 @@ services:
     env_file:
       - 'localstack/aws.env'
     environment:
-      DEBUG: ${DEBUG:-1}
+      DEBUG: ${DEBUG:-0}
       LS_LOG: WARN
-      SERVICES: s3,sqs,sns,firehose
+      SERVICES: s3,sqs,sns
```
alanplatt, line 307: is firehose not used for audit?

compose.land-grants.yml
```diff
@@ -10,6 +10,9 @@ services:
       - ./localstack/start-localstack-land-grants.sh:/etc/localstack/init/ready.d/start-localstack-land-grants.sh
 
   grants-ui:
+    depends_on:
+      land-grants-backend:
+        condition: service_healthy
```
DavidBarker-EE, line 15: [codex review]

**[P1] grants-ui waits for a healthcheck that does not exist**

grants-ui now depends on land-grants-backend with condition: service_healthy, but land-grants-backend has no healthcheck defined. I rendered the merged land-grants compose config and confirmed the backend service still has no healthcheck, while grants-ui waits on it as healthy.

That means the land-grants local stack can fail or hang when starting, especially with docker compose up --wait. Either change this back to service_started, or add a real healthcheck to land-grants-backend.

alanplatt (dismissed):

compose.land-grants.yml
davidatdefra: Fixed

alanplatt (dismissed):

alanplatt (approved):

PR #999, opened 2026-06-26
Use PrintSubmittedApplicationController that wires into dxt lifecycle instead of using the print submitted application hapi route handler

Andrew-Folga:


src/server/print-submitted-application/print-submitted-application.controller.js
```diff
@@ -1,189 +1,145 @@
...
   }
+
+  /**
+   * Gets the path to the status page (in this case /print-submitted-application page) for the GET handler.
+   * @param {object} [request] - The request object containing the URL info
+   * @param {object} [context] - The context object which may contain form state
+   * @returns {string} path to the status page
+   */
+  getStatusPath(request, context) {
+    return getPrintSubmittedApplicationPath(request, context, 'ConfirmationController')
```
dcapilla: shouldn't this be PrintSubmittedApplicationController?

src/server/common/helpers/print-application-service/print-application-service.js
```diff
@@ -173,8 +173,8 @@ export function buildPrintViewModel({
   applicantDetailsSections
 }) {
   return {
-    pageTitle: `${form.title} application`,
-    serviceName: form.title,
+    pageTitle: `${form.name} application`,
+    serviceName: form.name,
```
dcapilla, line 177: name can be undefined according to the typedef above, are we sure we want to do this without a fallback?

src/server/print-submitted-application/print-submitted-application.controller.js
```diff
@@ -1,189 +1,145 @@
...
+      const state = await cacheService.getState(request)
+
+      if (!state || state?.applicationStatus !== ApplicationStatus.SUBMITTED) {
+        return h.response('Application not submitted').code(statusCodes.forbidden)
+      }
+
+      const form = this.model.def
+      const slug = request.params.slug
+
+      return this.buildPrintResponse(
```
dcapilla: no try catch on this method to handle errors and log them

src/server/print-submitted-application/print-submitted-application.controller.js
Andrew-Folga: Good catch - fixing now

src/server/common/helpers/print-application-service/print-application-service.js
Andrew-Folga: Name should always be there if provided and I will amend the typedef above.

src/server/print-submitted-application/print-submitted-application.controller.js
Andrew-Folga: sorted ta

dcapilla (approved):

PR #1003, opened 2026-06-29
TGC-1374 Better error handling on example grant with map 

dcapilla:


src/server/common/map/map.plugin.js
```diff
@@ -33,8 +33,12 @@ async function parcelsHandler(request, h) {
         )
       )
     )
-  } catch {
-    return h.response({ error: 'unavailable' }).code(statusCodes.serviceUnavailable)
+  } catch (error) {
+    const message = /** @type {Error} */ (error).message
```
Andrew-Folga, line 37: do we want to log anything for monitoring as well?

Andrew-Folga (approved):
LGTM one comment

src/server/common/map/map.plugin.js
dcapilla: logging already happens further upstream when calling the fetch on land grants api, so adding it here would be redundant

PR #1005, opened 2026-06-29
Enhance getSafeRedirect function to reject protocol-relative URLs and…

alanplatt:
… add corresponding tests


src/server/auth/get-safe-redirect.js
```diff
@@ -1,10 +1,15 @@
...
+ * A leading `//` produces a protocol-relative URL that browsers treat as an
+ * absolute external destination, so it must be rejected even though it starts
+ * with '/'.
+ *
  * @param {string | null | undefined} redirect
  * @returns {string}
  */
 function getSafeRedirect(redirect) {
-  if (!redirect?.startsWith('/')) {
+  if (!redirect?.startsWith('/') || redirect.startsWith('//')) {
```
DavidBarker-EE: [P1] Backslash protocol-relative redirects still pass

The new guard blocks //evil.test, but it still allows /\evil.test. Browsers and the WHATWG URL parser normalise that to https://evil.test/, so this remains an open redirect variant. I confirmed the PR helper returns "/\\evil.test" unchanged, and new URL('/\\evil.test', 'https://service.gov.uk').origin resolves to https://evil.test.

I’d reject both protocol-relative forms, for example by requiring the second character to be neither / nor \, and add tests for /\\example.com and probably ///example.com.

alanplatt: Pushed a fix

DavidBarker-EE (approved):

PR #1021, opened 2026-07-02
TGC-1374 - Fix: backend sourced form definitions don't inherit shared redirect rules

dcapilla:
**Problem**: 

`grantRedirectRules: null` on form definitions is supposed to fall back to shared-redirect-rules.yaml defaults. That merge only lived in registerYamlForms (startup, YAML forms). Backend-sourced forms (BACKEND_FORM_DEF_ENABLED_SLUGS) skip that path entirely — their definition is fetched per-request from grants-ui-backend via resolveBackendDefinition, which never merged shared rules. Result: null reached forms-status-redirect.js, crashing on .preSubmission[0].

**Fix:**

- `resolveBackendDefinition` now applies the same shared-rules merge, per request instead of once at startup.
- Restored the `grantRedirectRules?.preSubmission?.[0]` null-check in forms-status-redirect.js as a safety net.
- Added a regression test for a backend-sourced form with `grantRedirectRules: null`.

### DIFF
--- a/src/server/common/forms/definitions/example-grant-with-map.yaml
+++ b/src/server/common/forms/definitions/example-grant-with-map.yaml
@@ -10,7 +10,9 @@ metadata:
     serviceName: Farm and land service
     cookiePolicyUrl: /cookies
     expiryDays: 365
-  grantRedirectRules: null
+  grantRedirectRules:
+    preSubmission:
+      - toPath: /summary
   permissions:
     enforce: false
   confirmationContent:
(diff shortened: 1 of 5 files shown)


src/server/common/forms/definitions/example-grant-with-map.yaml
```diff
@@ -10,7 +10,9 @@ metadata:
     serviceName: Farm and land service
     cookiePolicyUrl: /cookies
     expiryDays: 365
-  grantRedirectRules: null
+  grantRedirectRules:
```
davidatdefra: This is the default in the `shared-redirect-rules.yaml`, so should have already been being used?

dcapilla: it seems is not being used by example-grant-with-map as it is erroring on redirection. Will double check this

davidatdefra (approved):

PR #1024, opened 2026-07-02
TGC-1418 - Switching to OS Maps

dcapilla:


swdpcomputing (approved):

src/client/javascripts/parcel-map/index.js
```diff
@@ -186,80 +222,109 @@ class ParcelMap extends HTMLElement {
...
+  async #fetchData() {
+    /** @type {unknown} */
+    let lastError
+    for (let attempt = 0; attempt < FETCH_MAX_ATTEMPTS; attempt++) {
+      if (attempt > 0) {
+        await new Promise((resolve) => globalThis.setTimeout(resolve, FETCH_RETRY_DELAY_MS))
+      }
+      try {
+        const resp = await fetch(PARCELS_API_URL)
+        if (resp.ok) {
```
DavidBarker-EE, line 288: If the response isn't JSON we would still allow it through here. If we received a session error or any malformed upstream response it will skip through this try {} catch {}. In the old implementation we had `const body = await resp.json()` inside the try catch that would have caught this.

PR #1030, opened 2026-07-04
Refactor and extend audit coverage in acceptance tests

ScenarioFactory:
- Refactor approach to audit assertions to allow multiple concurrent scenarios to subscribe to the SNS topic
- Extend coverage across multiple scenarios
- Rename `whitelist.feature` to `allowlist.feature`

### DIFF
--- a/acceptance/test/features/user-permissions.feature
+++ b/acceptance/test/features/user-permissions.feature
@@ -23,6 +23,7 @@ Feature: User Permissions
         Given the user navigates to "/example-grant-with-auth"
         And logs in as CRN "1062311183"
         Then the user should see heading "You do not have permission to view this page"
+        And an unauthorised audit event should be published for grant "example-grant-with-auth" with CRN "1062311183" and SBI "106238911" and reason "permission"
 
         # unlock from CRN 1062311183
         Given there is no application lock for CRN "1062311183" and SBI "106238911" and grant "example-grant-with-auth"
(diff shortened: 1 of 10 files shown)


acceptance/test/features/user-permissions.feature
```diff
@@ -23,6 +23,7 @@ Feature: User Permissions
         Given the user navigates to "/example-grant-with-auth"
         And logs in as CRN "1062311183"
         Then the user should see heading "You do not have permission to view this page"
+        And an unauthorised audit event should be published for grant "example-grant-with-auth" with CRN "1062311183" and SBI "106238911" and reason "permission"
```
swdpcomputing, line 26: Does the app actually publish an unauthorised audit event with reason "permission"?

acceptance/test/support/world.js
```diff
@@ -39,8 +40,10 @@ setWorldConstructor(GrantsUiWorld)
 
 Before(async function () {
   await this.init()
+  this.auditQueue = await createScenarioAuditQueue()
 })
 
 After(async function () {
```
swdpcomputing, line 46: If createScenarioAuditQueue() or this.init()throws, this.auditQueue stays undefined, and then the After hook calls deleteScenarioAuditQueue(this.auditQueue) which destructures { queueUrl, subscriptionArn } off undefined and throws a TypeError whch masks the fail and leaks the SQS queue.

acceptance/test/features/user-permissions.feature
ScenarioFactory: It certainly does if the user is denied by virtue of user permissions.

acceptance/test/support/world.js
ScenarioFactory: Added guard to After function. The scenario will fail if the Before throws an error, but the After will still run and compound the error.

swdpcomputing (approved):

PR #1044, opened 2026-07-08
allow the SFD update URL in the CSP form-action directive so the Wood…

swdpcomputing:
…land redirect is not blocked

### DIFF
--- a/src/plugins/content-security-policy.js
+++ b/src/plugins/content-security-policy.js
@@ -1,5 +1,6 @@
 import { randomBytes } from 'node:crypto'
 import { config } from '~/src/config/config.js'
+import { error, LogCodes } from '~/src/server/common/helpers/logging/log.js'
 
 const defaultContentPolicy = (/** @type {string} */ nonce) => {
   const gtm = 'https://www.googletagmanager.com'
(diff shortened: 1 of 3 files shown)


src/plugins/content-security-policy.js
```diff
@@ -19,12 +20,29 @@ const defaultContentPolicy = (/** @type {string} */ nonce) => {
...
+    if (sfdUpdateUrl) {
+      let sfdUpdateUrlInvalid = false
+      try {
+        formActionSrc.push(new URL(sfdUpdateUrl).origin)
+      } catch {
+        sfdUpdateUrlInvalid = true
+      }
+
+      if (sfdUpdateUrlInvalid) {
+        log(LogCodes.SYSTEM.CSP_SFD_UPDATE_URL_INVALID, { sfdUpdateUrl })
```
davidatdefra: Rather than introduce a new variable `sfdUpdateUrlInvalid` you could just put the log inside the catch

src/server/common/helpers/logging/log-codes/system.js
```diff
@@ -173,5 +173,10 @@ export const SYSTEM = {
     level: 'warn',
     messageFunc: (messageOptions) =>
       `ensureUpdateDetailsPage: model.pages is empty for grantCode=${messageOptions.grantCode || 'unknown'} — pages may not have been initialised yet. If the forms engine has changed to async page initialisation, the queueMicrotask timing assumption no longer holds.`
+  },
+  CSP_SFD_UPDATE_URL_INVALID: {
+    level: 'warn',
```
davidatdefra: Probably should be `error` as it prevents the SFD redirect working.

swdpcomputing: Addressed.

src/plugins/content-security-policy.js
swdpcomputing: Addressed.

davidatdefra (approved):

PR #1045, opened 2026-07-08
Add handling of the Your draft application has been deleted to the frontend

Andrew-Folga:


src/server/application-deleted/views/application-deleted.njk
```diff
@@ -0,0 +1,42 @@
...
+    <h1 class="govuk-heading-xl">
+      Your draft application has been deleted
+    </h1>
+
+    <p class="govuk-body-l">
+      A change has been made to the service requiring all draft applications to be removed.
+    </p>
+
+    <p class="govuk-body">
+      <a
```
dcapilla: here you can use nunjucks, no need to have that svg data there

```
{% from "govuk/components/button/macro.njk" import govukButton %}

{{ govukButton({
  text: "Start a new application",
  href: "/{{ slug }}/start"
  isStartButton: true
}) }}
```

src/server/application-deleted/application-deleted.route.js
```diff
@@ -0,0 +1,31 @@
...
+      await cacheService.clearState(
+        /** @type {import('@defra/forms-engine-plugin/engine/types.js').AnyFormRequest} */ (
+          /** @type {unknown} */ (request)
+        ),
+        true
+      )
+    }
+
+    return h.view('application-deleted', {
+      slug: request.params.slug,
```
dcapilla, line 47: do we need to do this? I wonder if we already have access to the url slug on the nunjucks template. It's ok if we have no choice but seems redundant

src/server/application-deleted/views/application-deleted.njk
Andrew-Folga: Good point. Addressed now.

src/server/application-deleted/application-deleted.route.js
Andrew-Folga: @dcapilla sorry I was not able to find it

src/server/application-deleted/application-deleted.route.test.js
```diff
@@ -0,0 +1,116 @@
...
+      },
+      server: {}
+    }
+
+    const view = vi.fn()
+
+    const h = { view }
+
+    getState.mockResolvedValue({
+      applicationStatus: 'PURGED'
```
swdpcomputing, line 72: Magic string?

src/server/application-deleted/application-deleted.route.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @satisfies {ServerRoute}
+ */
+export const applicationDeletedRoute = {
+  method: 'GET',
+  path: '/{slug}/application-deleted',
+  handler: async (request, h) => {
+    const cacheService = getFormsCacheService(request.server)
+    const state = await cacheService.getState(request)
```
swdpcomputing: Can this throw? If so we won't see the page.

src/server/application-deleted/application-deleted.route.js
Andrew-Folga: in fact a few other pages are setting up the slug as well

DavidBarker-EE (dismissed):

src/server/application-deleted/application-deleted.route.test.js
Andrew-Folga: Good point - corrected now.

src/server/application-deleted/application-deleted.route.js
Andrew-Folga: Good point - addressed now.

dcapilla (approved):

DavidBarker-EE (approved):

swdpcomputing (approved):

PR #1051, opened 2026-07-09
Use named constant instead of string + land on slug instead of hardcoded /start page once link to start new application clicked

Andrew-Folga:


.npmrc
```diff
@@ -1,5 +1,5 @@
 git-tag-version=false
 ignore-scripts=true
-min-release-age=7
+min-release-age=0
```
dcapilla: why are we doing this here?

davidatdefra: ~DEFRA software security guidelines. It's to mitigate against supply chain attacks.~

davidatdefra: I must be red/green colour blind or something. Ignore that comment. Should not be changing this.

Andrew-Folga: We're temporarily setting min-release-age=0 because the security fix we need (body-parser) was published within the last 7 days.

Andrew-Folga: Normally we keep min-release-age=7 in line with DEFRA's supply-chain security guidance, which reduces exposure to newly published malicious or compromised packages. However, in this case the vulnerability scan is requiring a version that is newer than the configured release-age window allows npm to install.

Andrew-Folga: This change is intended only to unblock installation of the patched dependency. Once the package is older than the release-age threshold, we should revert to min-release-age=7.

swdpcomputing (dismissed):

DavidBarker-EE (approved):

.npmrc
Andrew-Folga: This is now reverted

swdpcomputing (approved):

dcapilla (approved):

PR #1053, opened 2026-07-09
TGC-1465: Parcel selection and actions config

davidatdefra:


src/server/land-grants/services/land-grants.service.js
```diff
@@ -78,15 +100,14 @@ export async function fetchAvailableActionsForParcel({ parcelId = '', sheetId =
...
+  const parcelIds = [parcelKey]
   const { parcels, groups: groupDefinitions = [] } = await parcelsWithExtendedInfo(parcelIds, LAND_GRANTS_API_URL)
   const foundParcel = parcels?.find((p) => p.parcelId === parcelId && p.sheetId === sheetId)
   const actionsForParcel = foundParcel?.actions?.map(mapAction) || []
 
-  const enabledActions = /** @type {string[]} */ (config.get('landGrants.enabledActions') ?? []).map((a) => a.trim())
   groupDefinitions.forEach((group) => {
     const groupActions = actionsForParcel.filter((a) => {
-      if (enabledActions.length > 0 && !enabledActions.includes(a.code)) {
+      if (!enabledActions.includes(a.code)) {
```
DavidBarker-EE, line 110: Wouldn't this mean that if `BACKEND_FORM_DEF_ENABLED_SLUGS` is empty this would now pass `[]` to the land actions controller without any meta data? Would this work? When this was empty (in the @UPL8 testing we did last week) there were no API actions available which was preventing any journey in CI from progressing. Shouldn't we add something to at least allow progression if there are no enabled features?

tools/apply-local-form-defs.mjs
```diff
@@ -0,0 +1,592 @@
...
+  }
+
+  const { overrides, warnings } = discoverOverrides()
+  for (const warning of warnings) {
+    console.warn(`  ⚠  ${warning}`)
+  }
+
+  if (!overrides.length) {
+    console.log('  No local form-definition overrides found — nothing to do.')
+    return 0
```
DavidBarker-EE: One from Codex

Remove stale overrides after their files are deleted — /tools/apply-local-form-defs.mjs:567-569

If a developer deletes or moves the last override YAML before toggling overrides off, discoverOverrides() returns an empty list and disable exits successfully without touching Mongo. The bumped config__form_definitions document then remains the highest active version in the persistent volume, even though the CLI reports overrides as disabled. Disable needs an independently persisted or queryable record of previously applied overrides so it can purge them without the source files.

src/server/common/map/map-select-page.controller.js
```diff
@@ -74,9 +80,18 @@ export default class MapSelectPageController extends withTaskContext(QuestionPag
 
     const selectedParcelsDisplay = selectedParcelIds.join(', ')
 
+    // In single-parcel-submission mode, selecting a parcel clears any previously selected parcel and its actions.
+    const clearedParcels = this.singleParcelSubmission ? { landParcels: {} } : {}
```
swdpcomputing, line 84: Does this remove all land parcels on every POST if there is no this.singleParcelSubmission? Is this desired?

src/server/land-grants/services/land-grants.service.js
davidatdefra: If the `BACKEND_FORM_DEF_ENABLED_SLUGS` env var is empty, that means there are no journeys full stop, so there would be no calls going through this service.
Basically what we're doing in this PR is moving the list of enabled land actions from a **global** env var to a grant-level metadata value. If you set that metadata value wrong or leave it blank, you will get issues like wrong actions or no actions (meaning you can't complete a journey), but it would only be for that grant, rather than all grants. We can't include any sort of "default" actions or a way through the journey where there are none, as it is a genuine requirement to have to select an action that is allowed for that grant before proceeding. We do need to be aware of which actions are needed for tests, and make sure the tests are testing reality, but in the case of UPL8 and UPL10 they aren't used in real journeys so the tests were not really valid.

src/server/common/map/map-select-page.controller.js
davidatdefra: No, we use `...clearedParcels` in the following lines to remove land parcels (by setting `landParcels: {}`) only if `singleParcelSubmission` is `true`
If it's `false` or `undefined` the destructuring is essentially a no-op - adding `...{}` to the state

alanplatt:
LGTM

DavidBarker-EE (approved):

PR #1056, opened 2026-07-10
TGC-1472 Moved SFD redirects out of check-details form submission to avoid SFD CSP hijacking

DavidBarker-EE:


src/server/sfd-redirect/index.js
```diff
@@ -0,0 +1,33 @@
+import { config } from '~/src/config/config.js'
+
+export const SFD_REDIRECT_SESSION_KEY = 'sfdRedirectUrl'
```
swdpcomputing, line 3: Should we have these in cdp-app-config?

DavidBarker-EE: I can't see any benefit to them being in CDP as they will never change across any environment, they're exported only because they're needed to marry up between the controller and this plugin.

swdpcomputing (approved):

PR #1063, opened 2026-07-14
TGC-1478 - Remove legacy yaml confing

dcapilla:


docs/ARCHITECTURE.md
```diff
@@ -272,17 +272,30 @@ For complete service configuration and setup, see [Docker Compose](./DOCKER.md#d
...
+resolved fresh per request from the backend's combined
+`POST /state/with-definition` endpoint (see
+[Forms Engine State Model](#forms-engine-state-model)), via
+`state-with-definition-context.js`'s `AsyncLocalStorage`-backed request
+context. Whitelist enforcement (`whitelist.js` `onPostAuth`) reads the grant's
+`whitelistCrnEnvVar`/`whitelistSbiEnvVar` from that same per-request envelope,
+not from the Redis entry, so it holds even on a fresh Redis or the first
+request after a publish.
+
+One consequence of resolving everything per-request: the dev-tools form picker (`getAvailableFormSlugs`/`getAllForms`) only lists slugs that have actually been visited at least once, not every grant that exists in the backend.
```
davidatdefra: If we're happy with losing the full list of grants from dev-tools, then I think we can get rid of Redis from involvement in caching forms stuff altogether. We already hold all the information we need via `request.app.model.def...` so we can delete `findFormBySlug` and tidy all that up.
Technically if we wanted to keep the full list of grants in dev-tools, we could get this by reading the list of yamls in the `config-broker-local` folder structure once at startup for local running only and add to a map or something to lookup from? Might be this is a later PR (which we'd need to create a ticket for)

davidatdefra:
`state-to-gas-payload-mapper.js` still has the legacy fallback 
```
const configVersion = request.app?.grantVersion ?? request?.app?.model?.def?.metadata?.version ?? '1.0.0'
```

src/server/common/forms/services/submission.js
```diff
@@ -52,14 +52,16 @@ export function loadSubmissionSchemaValidators() {
...
  * @param {object} payload - The submission answers to validate
  * @param {string} grantCode - The grant code whose schema to validate against
- * @returns {{valid: true, value: object} | {valid: false, errors: import('ajv').ErrorObject[]}} AJV validation result
+ * @param {string} schemaPath - `metadata.submission.submissionSchemaPath` from the resolved form definition
+ * @returns {{valid: true, value: object} | {valid: false, errors: import('ajv').ErrorObject[] | null | undefined}} AJV validation result
  */
-export function validateSubmissionAnswers(payload, grantCode) {
-  const validate = validators.get(grantCode)
-  if (!validate) {
+export function validateSubmissionAnswers(payload, grantCode, schemaPath) {
```
davidatdefra: This is only referenced in tests, we're not currently validating schemas, we've removed most of them from the yaml and repo anyway, and we intend to move to using a shared schema that comes from the backend/config broker at some point, so I would just remove all the schema loading and validating code for now.

docs/ARCHITECTURE.md
dcapilla: Good point, I think I am going to drop the list of grants from dev-tools and get rid of Redis involvement as part of this PR and create a JIRA ticket to add them later by reading the list of yamls

src/server/common/helpers/state/state-with-definition-context.js
```diff
@@ -137,7 +124,7 @@ export function getStateWithDefinition(request) {
 
     app.stateWithDefinition = fetchStateWithDefinitionFromApi(key, request, {
       lockToken: buildReadLockToken(request),
-      includeDefinition: isBackendSourcedSlug(grantCode)
+      includeDefinition: true
```
davidatdefra: I don't think we need `includeDefinition` at all now for `fetchStateWithDefinitionFromApi` as it was only used where a form came from legacy YAML file. So I'd update `fetch-saved-state-helper.js` to remove that and probably reword the code comments in that file.

src/server/common/helpers/whitelist/whitelist.js
```diff
@@ -11,6 +12,84 @@ export default {
   }
 }
 
+/**
```
davidatdefra: Just a heads up but my PR removes all the legacy whitelist code/references and does some renaming to allowlist so if that goes in first, there will be conflicts

davidatdefra (approved):

PR #1064, opened 2026-07-14
TGC-1472 SFD Redirect

DavidBarker-EE:


src/server/details-page/views/incorrect-details.njk
davidatdefra, line 1: Also need to make the same changes to `incorrect-details.html`
I believe the `.njk` version is needed for the dev tools journey runner.

davidatdefra (approved):

PR #1077, opened 2026-07-16
TGC-1477: Decommission whitelist

davidatdefra:


alanplatt (dismissed):

src/config/config.js
```diff
@@ -319,7 +319,7 @@ const convictConfig = {
       env: 'BACKEND_FORM_DEF_ENABLED_SLUGS'
     },
     backendAllowlistEnabledSlugs: {
```
dcapilla: if we are decommissioning whitelist, aren't we supposed to get rid of this config variable and always use allowlist from now on?

src/server/common/forms/definitions/example-whitelist.yaml
```diff
@@ -3,8 +3,6 @@ name: Example Whitelist
 metadata:
   id: 7bcafdd6-f8a7-4a31-8f98-8cd71e2c0f16
   referenceNumberPrefix: EW
-  whitelistCrnEnvVar: EXAMPLE_WHITELIST_CRNS
-  whitelistSbiEnvVar: EXAMPLE_WHITELIST_SBIS
```
dcapilla, line 7: we'll need to remove this from the config repo - these .yaml files are being removed entirely from here

dcapilla (approved):

PR #1087, opened 2026-07-20
decompose map plugin into route table, handlers and focused modules, …

swdpcomputing:
…quarantine mock, and reuse


src/server/common/map/handlers.js
```diff
@@ -0,0 +1,156 @@
...
+    logUpstreamError(
+      {
+        endpoint: ROUTES.parcelTiles,
+        service: SERVICE_LAND_GRANTS,
+        upstreamStatus: null,
+        errorMessage: parcelsResult.error.message
+      },
+      request
+    )
+    return h.response().code(statusCodes.serviceUnavailable)
```
dcapilla: this always returns 503 masking the actual error, perhaps we can try surfacing the actual error like we did on map.plugin.js (see `numericUpstreamStatus`)

src/server/common/map/handlers.js
```diff
@@ -0,0 +1,156 @@
...
+const SERVICE_LAND_GRANTS = 'land-grants-api'
+const CACHE_CONTROL_HEADER = 'Cache-Control'
+
+/**
+ * The land-grants client reports the upstream status as `code` or `status`
+ * depending on where the call failed.
+ * @param {Error & { code?: unknown, status?: unknown }} error
+ * @returns {number | undefined}
+ */
+function numericUpstreamStatus(error) {
```
dcapilla: I know you didn't introduce this, but perhaps we can replace this with getUpstreamStatus?? They seem to be doing the same thing

dcapilla (approved):

PR #1090, opened 2026-07-21
harden map tile routes with clamped z/x/y schemas, generic parcel err…

swdpcomputing:
Clamp OS/parcel tile params to valid zoom and x/y ranges so a signed-in user can no longer script arbitrary coordinates against the metered OS key; return a generic parcels error instead of leaking the upstream body; guard a null OS tile body (502) and a missing OS_MAPS_API_KEY at startup; and gate CartoCDN in the CSP on devMode so it no longer ships on every page.

### DIFF
--- a/src/server/common/map/tile-params.js
+++ b/src/server/common/map/tile-params.js
@@ -1,0 +1,29 @@
+import Joi from 'joi'
+import { OS_MIN_ZOOM, OS_MAX_ZOOM } from './os-maps.js'
+
+/**
+ * Builds a per-route tile-param schema that clamps `z` to a valid zoom range and
+ * `x`/`y` to `[0, 2^z)` — the only valid tile coordinates at that zoom. The
+ * `x`/`y` bound depends on `z`, so it lives in an object-level `.custom()`; a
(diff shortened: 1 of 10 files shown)


src/server/common/map/tile-params.js
```diff
@@ -0,0 +1,29 @@
...
+      return value.x >= bound || value.y >= bound ? helpers.error('any.invalid') : value
+    })
+  }
+}
+
+// Two named schemas so each route validates its own range. They share the OS
+// zoom window today (same viewport), but the seam lets the parcel range diverge
+// without touching the OS route.
+export const osTileParams = makeTileParams({ minZoom: OS_MIN_ZOOM, maxZoom: OS_MAX_ZOOM })
+export const parcelTileParams = makeTileParams({ minZoom: OS_MIN_ZOOM, maxZoom: OS_MAX_ZOOM })
```
dcapilla, line 29: I know you didn't change this but OS_MAX_ZOOM here should be 22 and different than the one on the previous line - here's what the Land Grants API has set: 

https://github.com/DEFRA/land-grants-api/blob/vector-tile-spike/src/features/vector-tiles/schema/parcelTiles.schema.js#L3

dcapilla (approved):

PR #1103, opened 2026-07-23
TGC-1481: Forward Defra ID user context to Land Grants API

DavidBarker-EE:
- forwards the signed-in user's Defra ID token on every Land Grants API request using `x-forwarded-authorization`
- adds the authenticated SBI as a string only to `/api/v2/parcels`
- sends the authenticated SBI as a number to `/api/v2/application/validate
- sends the user context through Grasslands, map, payment, and woodland
- fails when the Defra ID token or SBI is missing (for routes that require it)


src/contracts/v2/land-grants.client.contract.test.js
```diff
@@ -761,8 +768,8 @@ describe('validate', () => {
       .withRequest({
         method: 'POST',
         path: '/api/v2/application/validate',
-        headers: { 'Content-Type': 'application/json' },
-        body: payload
+        headers: makeLandGrantsHeaders(),
+        body: withAuthenticatedSbi(payload)
```
giftig: Note also this is providing SBI as a string, but the endpoint currently wants it to be a number. We'd ideally like to change it to string like we've asked for on the `/v2/parcels` endpoint but we'll need to coordinate changes between both services to make that happen, and if you send it as a string right now we'll reject it.

So please continue to send it as a number on `/v2/application/validate` for the time being, but send it as a string on `/v2/parcels`.

DavidBarker-EE: No problem, I'll change that

src/server/land-grants/controllers/select-actions-base-page.controller.js
```diff
@@ -244,12 +249,15 @@ export default class SelectActionsBasePageController extends QuestionPageWithPar
...
 
     try {
-      const validationResult = await validateApplication({
-        applicationId: /** @type {string} */ (referenceNumber),
-        sbi: /** @type {string} */ (sbi),
-        crn: /** @type {string} */ (crn),
-        state
-      })
+      const userContext = getLandGrantsUserContext(request)
+      const validationResult = await validateApplication(
```
dcapilla, line 253: Wouldn't it be easier to pass `request` to these methods and get the userContext inside `validateApplication`, `fetchAvailableActionsForParcel`, etc etc so we avoid calling getLandGrantsUserContext every time we call one of those methods in the future?

I am fine with this if you don't want to change its current shape.

src/server/land-grants/services/land-grants.client.js
```diff
@@ -117,82 +125,92 @@ export async function fetchParcelTile(parcelIds, z, x, y, baseUrl) {
...
+ * @param {LandGrantsUserContext} userContext
  * @returns {Promise<ValidateApplicationResponse>} - Validation result
  * @throws {Error}
  */
-export async function validate(request, baseUrl) {
+export async function validate(request, baseUrl, userContext) {
+  const { sbi } = validateLandGrantsUserContext(userContext)
   const endpoint = '/api/v2/application/validate'
-  return postToLandGrantsApi(endpoint, request, baseUrl)
+  return postToLandGrantsApi(endpoint, { ...request, sbi: Number(sbi) }, baseUrl, userContext)
```
dcapilla, line 197: are we supposed to pass the sbi to the `/validate` endpoint? I thought we were just passing it for `/v2/parcels`

giftig: @dcapilla yeah you've always been passing it to `/application/validate`, that hasn't changed here, it's just been refactored a bit.

DavidBarker-EE: Rob Moore requested I send it as an integer to `validate` and as a string to `/v2/parcels`

src/server/land-grants/controllers/select-actions-base-page.controller.js
DavidBarker-EE: I didn't really want to couple the services to the framework, it would probably be more simplistic but I tend to always go for loose coupling. I don't mind tbh but I think this way is more re-usable (not that it probably matters)

src/server/land-grants/services/land-grants.client.js
dcapilla: ah sorry, I didn't remember that! Thanks

dcapilla (approved):

PR #1117, opened 2026-07-28
mask CRN to last 4 digits and rename user log field to CRN in auth logs

swdpcomputing:
Add a maskCrn helper that reveals only the last 4 digits of a CRN, and rename the auth log field from user/userCRN to CRN across all AUTH log codes. This stops CDP obfuscating the value (it was masked because labelled "user") while avoiding logging the full CRN as PII, aiding live support triage of unauthorised access attempts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,8 @@
     "joi": "17.13.4",
     "body-parser": "1.20.6",
     "brace-expansion": "2.1.2",
-    "postcss": "8.5.19"
+    "postcss": "8.5.19",
+    "ip-address": "10.2.1"
   },
   "dependencies": {
     "@aws-sdk/client-sns": "3.1091.0",
--- a/src/server/common/helpers/logging/log-codes/auth.js
+++ b/src/server/common/helpers/logging/log-codes/auth.js
@@ -1,3 +1,5 @@
+import { maskCrn } from '~/src/server/common/helpers/logging/mask-crn.js'
+
 /**
  * @type {Object<string, import('./definition.js').LogCodesDefinition>}
  */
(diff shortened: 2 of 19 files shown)


src/server/common/helpers/logging/log.test.js
```diff
@@ -64,7 +64,7 @@ describe('Logger Functionality', () => {
 
     log(LogCodes.AUTH.SIGN_IN_SUCCESS, testOptions)
 
-    expect(logger.info).toHaveBeenCalledWith({}, 'User sign-in successful for user=test-user, organisation=test-org')
+    expect(logger.info).toHaveBeenCalledWith({}, 'User sign-in successful for CRN=test-user, organisation=test-org')
```
davidatdefra: Very minor: might be helpful to use real-looking (maskable) CRNs in these tests to again prove the masking works

src/server/common/helpers/logging/log-codes/auth.js
davidatdefra, line 1: All good here, but other log codes still output messageOptions.userId as `user` unmasked: land-grants.js, declaration.js, confirmation.js, agreements.js, forms.js, print-application.js, permissions.js, system.js, and resource-not-found.js

package.json
```diff
@@ -97,7 +97,8 @@
     "joi": "17.13.4",
     "body-parser": "1.20.6",
     "brace-expansion": "2.1.2",
-    "postcss": "8.5.19"
+    "postcss": "8.5.19",
+    "ip-address": "10.2.1"
```
davidatdefra, line 101: There is a .snyk ignore for this as the fixed version is less than 7 days old. If it is a high severity and affects the runtime code, fine to add the override, but should remove the ignore.

davidatdefra (approved):

PR #1118, opened 2026-07-28
TGC-1482 - Further UX changes and improvements on actions page

dcapilla:


src/server/land-grants/controllers/select-actions-page.controller.js
```diff
@@ -74,6 +106,18 @@ export default class SelectActionsPageController extends SelectActionsBasePageCo
     return getSelectedActionCodes(payload)
   }
 
+  /**
+   * @param {object} payload
+   * @param {Array} actions
+   * @returns {Array<{ code: string, description: string, value?: string|number }>}
+   */
+  getAddedActionsForValidationError(payload, actions) {
```
swdpcomputing, line 114: should this fall back o getAddedActionsFromPayload like the quantity-error path does?

swdpcomputing (approved):

src/server/land-grants/controllers/select-actions-page.controller.js
dcapilla: No, I think this is fine. 

getAddedActionsFromPayload's fallback reads from state, so that's the last saved value. But there's a problem: when two non-quantity actions (CMOR1 + CLIG3) get chosen and compete for the same land in one submission, we can hit a validation error. In that case we want the exact snapshot of actions from right before submission, even if their values are invalid, this is so we can display the error and pre-populate the page with the same values it had when it failed (so we don't lose data). That's what plannedActionsSnapshot (written client-side after every successful live refresh) gives us, which state can't.

PR #1119, opened 2026-07-29
TGC-1417: Toggle brand assets by grant assessment phase

DavidBarker-EE:


davidatdefra (dismissed):
Sonar is having a little whinge about `src/config/nunjucks/context/build-service-branding.js` but otherwise looks good to me

davidatdefra (approved):

PR #1138, opened 2026-08-05
TGC-1503 - Mapping enhancements

dcapilla:
Mapping enhancements:

- Select your land parcel now displays farm name, total size and number of parcels on the right hand side
- Page width has been increased according to prototype
- When a parcel is selected, a new table will appear below the map to display parcel information
- When a parcel has been selected and user clicks "Change", parcel becomes deselected
- We've refactored the way we fetch mock data, we spotted a race condition with the map loading faster than a `yar.set` and map was failing to load, so we have eliminated that yar.set logic for a simpler logic that doesn't have race conditions.


acceptance/test/steps/when.steps.js
```diff
@@ -6,20 +6,47 @@ import AutocompleteField from '../page-objects/auto-complete.field.js'
...
 import MonthYearField from '../page-objects/month-year.field.js'
 
+When('the map has loaded parcel {string}', async function (parcelId) {
+  // Real usage populates the map totals from the parcel-map:ready event, which the
+  // <parcel-map> element only dispatches once MapLibre has actually loaded (a
+  // working WebGL context)
+  await this.page.waitForLoadState('domcontentloaded')
+  await this.page.evaluate((id) => {
+    document.getElementById('parcel-map').dispatchEvent(
+      new CustomEvent('parcel-map:ready', {
```
ScenarioFactory, line 16: Why do we need to send this, won't the page itself do that as the map loads?

acceptance/test/features/mapping.feature
```diff
@@ -11,6 +18,8 @@ Feature: Mapping
 
         # select-land-parcel
         Then the user should be at URL "select-land-parcel"
+        When the map has loaded parcel "SD6352-8774"
+        And the user should see the available land parcels total populated
```
ScenarioFactory: *Then

acceptance/test/steps/when.steps.js
dcapilla: For the same reason we were already doing the same event triggering on "the user selects parcel {string} on the map" step

I added a comment at the top of mapping.feature explaining it

```
    # "the map has loaded parcel", "selects parcel ... on the map" and "clicks
    # 'Change' on the selected parcel summary" fake the parcel-map:ready/
    # parcel-map:selection CustomEvents directly, rather than driving the real
    # <parcel-map> web component - headless CI Chromium has no working WebGL,
    # so MapLibre never actually initialises there. This proves the page's
    # event-handling wiring but not that a user can genuinely see and click a
    # rendered map. Revisit once CI has a way to render WebGL and interact with the actual map.
    ```
    
 we need to change the way we're running this test as it is not representing reality - but I didn't want to do that as part of this mapping enhancements because it is not something we introduced here

davidatdefra (dismissed):

davidatdefra (approved):

PR #1145, opened 2026-08-07
WIP - add a remove land parcel confirmation page and clear map selection state when a parcel is removed.

swdpcomputing:
Adds a "Remove this land parcel?" confirmation page for grasslands (confirm button plus cancel link, rather than the yes/no radios used for action removal) and flashes the removed parcel reference so the "Your land and actions" page can show a success banner. Makes RemoveActionPageController's previously hardcoded redirect paths configurable per grant via YAML page config, defaulting to the existing farm payments paths so that journey is unchanged. Also clears the map's separate selection keys when a parcel leaves state, which previously left a removed parcel highlighted on the map and listed on Check your answers.

### DIFF
--- a/src/server/land-grants/view-state/land-parcel.view-state.js
+++ b/src/server/land-grants/view-state/land-parcel.view-state.js
@@ -233,6 +233,41 @@ export function getAddedActionsForStateParcel(state, selectedLandParcel) {
 }
 
 /**
+ * Drops a parcel from the map's own record of what is selected, which
+ * MapSelectPageController keeps separately from `landParcels`. Without this a
+ * removed parcel stays highlighted on the map and keeps appearing on Check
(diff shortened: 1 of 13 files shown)


src/server/land-grants/view-state/land-parcel.view-state.js
```diff
@@ -269,6 +305,7 @@ export function deleteActionFromState(state, parcel, action) {
     // Remove parcel if no actions remain
     if (Object.keys(newState.landParcels[parcel].actionsObj).length === 0) {
       delete newState.landParcels[parcel]
+      pruneMapSelection(newState, parcel)
```
DavidBarker-EE, line 308: Do we want to prune the map selection here? IIRC `deleteActionFromState` uses this and will add an unwanted side effect for that journey that asks them to select actions for the same selected parcel.

PR #1171, opened 2026-08-19
Adopt the parcels API availability and inputRequired contract

swdpcomputing:
The Land Grants parcels API replaced `action.availableArea` with `action.availability` (`0` means not compatible, `null` means no restriction), driven by actions now being measured in linear metres as well as area, caught by the PACT test. Migrates the type definitions, service, view model, view state, validator, controllers, plugin and client script to the new shape, adds shared `getAvailabilityLimit`/`formatUnit`/`UNIT_TYPES` helpers, and widens the planned-actions unit allowlist beyond `ha`/`sqm`. `inputRequired` is adopted at the same time because the quantity-input decision previously read `availability.type`, the same key the API repurposes.

### DIFF
--- a/mockserver/expectations.json
+++ b/mockserver/expectations.json
@@ -188,7 +188,7 @@
             "actions": [
               {
                 "code": "TEST1",
-                "availableArea": {
+                "availability": {
                   "value": 7.0,
                   "unit": "ha"
                 },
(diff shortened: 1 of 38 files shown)


Andrew-Folga (dismissed):
LGTM Just a few comments I posted about the hardcoded `ha` values on the payment-rate copy.

davidatdefra (approved):

Andrew-Folga (approved):
LGTM

PR #1175, opened 2026-08-20
Correct land grants and claims types without exposing type debt

swdpcomputing:
Replaces generic {object} JSDoc across the land grants mappers, services and payload typedefs with concrete types (ParcelItem, AgreementLevelItem, PaymentCalculation, ValidateApplicationResponse), and fixes the two defects those types surfaced: upsertCurrentClaim no longer overwrites stored totals with undefined, and calculateLandActionsPayment now returns an error message instead of formatting NaN when the annual total is missing. Shared land grants test fixtures move into src/mocks/land-grants-mocks.js. Carries none of the @ts-nocheck suppressions or tsconfig changes from the originating branches - tsc, eslint and the full 3400-test suite pass unchanged.

### DIFF
--- a/src/server/claims/services/claim-state.js
+++ b/src/server/claims/services/claim-state.js
@@ -80,7 +80,12 @@ export function upsertCurrentClaim(state, { referenceNumber, totalEligibleArea, 
   const currentIndex = claims.findIndex((claim) => claim?.status !== ClaimStatus.SUBMITTED)
 
   if (currentIndex >= 0) {
-    claims[currentIndex] = { ...claims[currentIndex], totalEligibleArea, unit, totalClaimAmountPence }
+    claims[currentIndex] = {
+      ...claims[currentIndex],
+      ...(totalEligibleArea !== undefined && { totalEligibleArea }),
+      ...(unit !== undefined && { unit }),
+      ...(totalClaimAmountPence !== undefined && { totalClaimAmountPence })
+    }
     return { claims, currentClaim: claims[currentIndex] }
   }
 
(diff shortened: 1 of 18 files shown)


DavidBarker-EE (dismissed):

src/server/claims/services/claim-state.js
```diff
@@ -80,7 +80,12 @@ export function upsertCurrentClaim(state, { referenceNumber, totalEligibleArea,
   const currentIndex = claims.findIndex((claim) => claim?.status !== ClaimStatus.SUBMITTED)
 
   if (currentIndex >= 0) {
-    claims[currentIndex] = { ...claims[currentIndex], totalEligibleArea, unit, totalClaimAmountPence }
+    claims[currentIndex] = {
```
davidatdefra: Not sure about this.
There is a chance things can get out of sync here if there are different API failures on repeat visits, e.g. user goes to page, everything works, but for some reason they don't submit.... they come back and the area API call works, but the claim amount one doesn't, but the conditional destructuring means we keep the old (now invalid) amount and it gets sent to GAS.
Unlikely to happen but especially with rural internet, not impossible.
Would be better to make the updates atomic across all API calls

The previous version also had issues, but I'm hoping would fail validation at GAS as some fields would be undefined.

davidatdefra (changes requested):

src/server/claims/services/claim-state.js
davidatdefra: Note: we might need UX advice on what to show in the case of one or more API failures, but we should never keep stale data as the values can change in theory on each visit to the page.

davidatdefra (approved):

PR #1191, opened 2026-08-25
Extend claims acceptance test coverage to check GAS submission

ScenarioFactory:
- Extend claims coverage to check GAS submission
- Remove unnecessary guard for the presence of MockServer left over from time when these tests ran in CDP
- Refactor some inappropriate assertions on the results of `Array.prototype.find` where we could receive `undefined`

### DIFF
--- a/acceptance/test/steps/gas.steps.js
+++ b/acceptance/test/steps/gas.steps.js
@@ -12,18 +12,16 @@ const schemas = {
   'example-grant-with-auth': require('../../schemas/example-grant-with-auth-submission.schema.json')
 }
 
-if (process.env.MOCKSERVER_HOST) {
-  Before(async function () {
-    const expectationId = await Gas.setDefaultStatusQuery404Response()
-    expectationIds.push(expectationId)
-  })
+Before(async function () {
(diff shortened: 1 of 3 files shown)


acceptance/test/steps/gas.steps.js
```diff
@@ -103,6 +101,23 @@ Then('the GAS submission should be valid against the {string} schema', async fun
...
 
+Then(
+  'the claim {string} for CRN {string} and SBI {string} should be submitted to GAS',
+  async function (claimNumberSuffix, crn, sbi) {
+    if (!referenceNumbers.current) {
+      throw new Error('No reference number stored by earlier step')
+    }
+
+    const request = await Gas.getClaimSubmission(referenceNumbers.current)
+    expect(request).not.toBeNull()
```
DavidBarker-EE: Minor: This check probably isn't what you want here, e.g. `expect(undefined).not.toBeNull()` will be true. Would be better to use `expect(request).toBeDefined()`

ScenarioFactory: Agreed. Fixing.

DavidBarker-EE (approved):

PR #1194, opened 2026-08-25
Bugfix/tgc 1611 restore rejected parcel selection

swdpcomputing:


package.json
```diff
@@ -74,6 +74,7 @@
   "author": "Defra DDTS",
   "license": "OGL-UK-3.0",
   "overrides": {
+    "fast-uri": "3.1.6",
```
davidatdefra: This version is newer than the `min-release-age=7` allows, and is a build/dev-tooling only, so has been added to `.snyk` ignore list with an expiry after the cooldown period. Need to revert this change and the lock file

src/server/common/map/map-select-page.controller.js
```diff
@@ -3,14 +3,16 @@ import { withTaskContext } from '~/src/server/task-list/task-list.helper.js'
...
+import { formatParcelReference, parseLandParcel } from '~/src/shared/format-parcel.js'
+import { escapeHtml } from '~/src/server/common/utils/escape-html.js'
 import { log, error, LogCodes } from '~/src/server/common/helpers/logging/log.js'
 import { isNoActionsMockEnabled } from '~/src/server/dev-tools/mock-overrides.js'
 
-// Must match the no-eligible-actions copy in
-// src/server/land-grants/views/select-actions.html
-const NO_ELIGIBLE_ACTIONS_ERROR =
-  'There are no eligible actions for this parcel.<br>' +
+// Second line must match the no-eligible-actions copy in
```
davidatdefra: Should we remove the "else" section from the select-actions.html template now that scenario is handled by redirecting back to the map?

davidatdefra (approved):

PR #1205, opened 2026-08-26
TGC-1535: GAS claim entitlement

davidatdefra:


src/server/common/services/grant-application/grant-application.service.js
```diff
@@ -307,6 +307,20 @@ export async function getApplicationStatus(code, clientRef, request) {
...
+/**
+ * Fetches the available claim entitlements for an application from GAS
+ * @param {string} code - Grant code
+ * @param {string} clientRef - Application client reference
+ * @param {AnyRequest | undefined} request - Request object
+ * @returns {Promise<any>} - Promise that resolves to the available claims payload JSON
+ * @throws {GrantApplicationServiceApiError} - If the API request fails
+ */
+export async function getAvailableClaimEntitlements(code, clientRef, request) {
+  const url = `${GAS_API_ENDPOINT}/grants/${code}/entitlements/${clientRef.toLowerCase()}/available-claims`
```
swdpcomputing: Is clientRef checked truthy somewhere else? If fail here would be a silent fail...

src/server/claims/controllers/start-claim-page.controller.js
```diff
@@ -79,15 +81,26 @@ export default class StartClaimPageController extends QuestionPageController {
...
+   * `totalHectares.value` onto `totalEligibleArea`.
+   * @param {AnyFormRequest} request
+   * @param {FormContext} context
    * @returns {Promise<Record<string, string | number>>} data items keyed by item name
    */
-  async fetchGasEntitlements(_request, _context) {
+  async fetchGasEntitlements(request, context) {
+    const grantCode = getGrantCode(request)
+    const state = context.state ?? {}
+    const clientRef = /** @type {string} */ (state.$$__referenceNumber)
```
swdpcomputing: Do we need to check __referenceNumber validity?

src/server/common/services/grant-application/grant-application.service.js
davidatdefra: We know `clientRef` cannot be `null` as it comes from `state.$$__referenceNumber` which also can never be `null`

src/server/claims/controllers/start-claim-page.controller.js
davidatdefra: We set that value and it never changes for the life of the grant application, so I don't think we need any other checks.

swdpcomputing (approved):

PR #1206, opened 2026-08-27
TGC-1619: Improve HTTP client onboarding

davidatdefra:


tools/init-http-client-secrets.js
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * The encrypted bearer tokens this tool can generate. Each entry maps a key in
+ * http-client.private.env.json to the raw token and encryption key used to
+ * build the encrypted bearer value the backend expects.
+ */
+const generatedTokens = [
+  {
+    key: 'brokerAuthToken',
+    token: process.env.CONFIG_BROKER_AUTH_TOKEN ?? 'config-broker-auth-token',
```
Andrew-Folga: Should these use GRANTS_CONFIG_BROKER_AUTH_TOKEN and GRANTS_CONFIG_BROKER_ENCRYPTION_KEY? Those are the variable names used by compose.grants-ui.yml and documented for this workflow. As written, configured broker credentials are ignored and the generated token uses the local fallback values.

davidatdefra: Good spot - have fixed, thanks.

Andrew-Folga (approved):
LGTM

PR #1209, opened 2026-08-28
TGC-1610: Multi claim GAS check

davidatdefra:


swdpcomputing (approved):
I feel that the jsdoc is somewhat verbose, but fine.

PR #1212, opened 2026-08-28
Consistent quantity validation for land actions

swdpcomputing:
Adds a shared client/server rule set for land action quantities that normalises bare decimals (".5" becomes "0.5") and reports distinct messages for non-numeric, zero or negative, over-precision and over-available values. The client now validates and renders each error in both the error summary and on the field, matching how the server already reports them, and the "X hectares available" text moves from the conditional panel into the checkbox hint.


acceptance/test/steps/action.steps.js
```diff
@@ -55,7 +55,12 @@ Then('(the user )should see {string} hectares available for action {string}', as
   await expect(this.page.locator(`#landActionQuantity_${action}-hint`)).toContainText(`${quantity} hectares available`)
 })
 
+Then('(the user )should see {string} in the quantity field for action {string}', async function (quantity, action) {
```
DavidBarker-EE: You'll need to run this by @ScenarioFactory

src/shared/action-quantity-validation.js
```diff
@@ -0,0 +1,75 @@
...
+
+export const QUANTITY_PRECISION = 4
+
+const PLAIN_DECIMAL = /^-?(\d+(\.\d*)?|\.\d+)$/
+
+export const QUANTITY_ERRORS = {
+  NOT_A_NUMBER: 'Must be numbers',
+  NOT_GREATER_THAN_ZERO: 'Value must be greater than 0',
+  TOO_MANY_DECIMAL_PLACES: `No more than ${QUANTITY_PRECISION} dp`,
+  MORE_THAN_AVAILABLE: 'More than available area'
```
DavidBarker-EE: **Major.** This table is meant to be the single source of truth, but two of these never reach the user the way they read here.

`MORE_THAN_AVAILABLE` is dead on the server: `validateSelectedActionQuantities` calls `getQuantityError(rawValue)` with no `max`, and the over-available scenario in `action-selection.feature` keeps its old expectation while every other quantity message in that file was rewritten. So the user sees `More than available area` on blur and the API's wording after pressing continue, for the same mistake. Either put the server and API copy here, or pass `max` through the validator so this table really is the source.

The copy itself also reads worse than what it replaces. `Must be numbers` is plural for a single value, `Value must be greater than 0` names nothing the user recognises, and `No more than 4 dp` abbreviates in a way that will not clear content review. They now sit next to `Enter a quantity for Herbal leys: CSAM3`, which is still generated a few lines away in the validator. Something like `The quantity must be a number`, `The quantity must be more than 0`, and `` `The quantity must have ${QUANTITY_PRECISION} decimal places or fewer` `` would read consistently. Changing them means updating the matching strings in `action-selection.feature`.

src/shared/action-quantity-validation.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/**
+ * Whether a quantity is usable as a claim.
+ * @param {string | number | null | undefined} raw
+ * @param {number} [max]
+ * @returns {boolean}
+ */
+export function isValidQuantity(raw, max) {
+  return String(raw ?? '').trim() !== '' && getQuantityError(raw, max) === null
```
DavidBarker-EE: **Minor.** `getQuantityError('')` returns `NOT_A_NUMBER`, which is the wrong message for a blank field, so the required-field rule ends up hand-written in both call sites: `rawValue === ''` in the validator and `quantityInput.value === '' ? null` in the client. Correct today, and a trap for the third consumer, who will render `Must be numbers` over an empty input.

A `REQUIRED` sentinel returned for empty input, or a JSDoc note that callers handle empty first, would prevent that. Note the `trim() !== ''` clause here is redundant either way, since empty already fails `PLAIN_DECIMAL`. Worth keeping as-is only if you want the two functions to stay independently readable.

src/client/javascripts/land-grants/select-actions-page.js
```diff
@@ -777,9 +688,12 @@ function bindQuantityFocusBlurHandlers(form, refreshAvailability) {
   form.addEventListener(
     'blur',
     (event) => {
-      if (getCheckboxForQuantityTarget(form, event.target)) {
-        flushPending()
+      const checkbox = getCheckboxForQuantityTarget(form, event.target)
+      if (!checkbox) {
+        return
       }
+      normaliseAndValidateQuantity(checkbox)
```
DavidBarker-EE: **Blocker.** Blur validates unconditionally, and the else branch of `normaliseAndValidateQuantity` clears the error whenever the client finds no fault, including the forced `value === '' ? null` case.

Sequence: the server rejects an empty quantity with `Enter a quantity for Herbal leys: CSAM3`, the page redisplays with the inline message and the summary entry, the user clicks into the field and clicks away without typing. No `input` event fires, so the `valueOnFocus` edit guard just above (line 672) never runs, and this blur path deletes the inline message, the summary entry and the summary itself while the problem still applies. Any API-side error the client cannot re-derive goes the same way.

`isProtectedFromRefresh(checkbox)` is already exported from `action-checkbox-state.js` for exactly this distinction, and the first real edit clears the flag, so gating on it here fixes the case:

```suggestion
      const checkbox = getCheckboxForQuantityTarget(form, event.target)
      if (!checkbox) {
        return
      }
      if (!isProtectedFromRefresh(checkbox)) {
        normaliseAndValidateQuantity(checkbox)
      }
      flushPending()
```

A narrower fix is to return early from `normaliseAndValidateQuantity` on an empty normalised value instead of clearing.

src/client/javascripts/land-grants/select-actions-page.js
```diff
@@ -127,92 +89,37 @@ function disableOtherActions(form, triggeringCheckbox) {
...
-  const field = getChosenAreaField(checkbox)
-  if (field) {
-    field.value = '0'
+  quantityInput.value = normaliseQuantityInput(quantityInput.value)
+  const message =
+    quantityInput.value === '' ? null : getQuantityError(quantityInput.value, getTotalAvailableArea(checkbox))
+  if (message) {
+    showQuantityError(quantityInput, message)
+  } else {
+    clearQuantityError(quantityInput)
```
DavidBarker-EE: See the blocker on the blur handler below. The `else` here is what erases a server-rendered error when the client finds nothing wrong with the current value, and the empty case is forced down that branch by the ternary above it.

Normalising on blur rather than on `input` is the right call, by the way. No caret fighting while someone types `0.`.

src/server/land-grants/view-models/select-actions.view-model.js
```diff
@@ -103,7 +102,7 @@ function getHintHtml(action, needsQuantity) {
   const rateText = `Payment rate per year: £${action.ratePerUnitGbp?.toFixed(2)}/ha${agreementRateText}${requirementLineText}`
   const limit = getAvailabilityLimit(action.availability)
   const availabilityHintHtml =
-    !needsQuantity && limit != null
+    limit != null
       ? `<br><span id="${getActionQuantityFieldName(action.code)}-hint">${limit} ${formatUnit(action.availability?.unit)} available</span>`
```
DavidBarker-EE: Question. Widening the condition moves a quantity action's availability text out of the njk `hint.text`, which `govukInput` escaped and this PR removes, into this concatenated string. `limit` is numeric, but is `unit` escaped on this path? The changed lines do not show it, and the same expression already existed for non-quantity actions, so this may be pre-existing rather than new.

The hint id and the input's `describedBy` are sound, since both derive from this same `limit != null` condition and so cannot dangle.

src/server/land-grants/controllers/select-actions-page.controller.test.js
```diff
@@ -410,29 +418,51 @@ describe('SelectActionsPageController', () => {
...
     })
 
+    test('should normalise a bare decimal quantity to a leading-zero value before saving state', async () => {
+      mockRequest.payload = { landAction: 'UPL2', landActionQuantity_UPL2: '.5' }
+
+      await post()
+
+      expect(controller.setState).toHaveBeenCalled()
+      const stateArg = controller.setState.mock.calls[0][1]
+      expect(stateArg.landParcels['sheet1-parcel1'].actionsObj.UPL2.value).toBe(0.5)
```
DavidBarker-EE: **Minor.** This assertion cannot fail if `normaliseQuantityFields` is deleted: `Number('.5')` and `Number('0.5')` are the same value, and `getQuantityError` normalises internally, so `.5` passes validation unaided.

The one observable server-side effect of the new normalisation is the value redisplayed in the quantity input after a validation error, through `getAddedActionsForValidationError`. Submitting `.51234` and asserting both the decimal-places error and a redisplayed `0.51234` would cover it.

src/shared/action-quantity-validation.test.js
```diff
@@ -0,0 +1,79 @@
...
+    expect(getQuantityError(raw, AVAILABLE)).toBe(expected)
+  })
+
+  it.each([['11.2200'], ['11.22'], ['10'], ['.5'], ['0.5'], ['0.0001']])('accepts %j', (raw) => {
+    expect(getQuantityError(raw, AVAILABLE)).toBeNull()
+  })
+
+  it.each([['1e5'], ['14.211.442121'], ['0x10'], ['Infinity'], ['5 ha'], ['']])('rejects %j as not a number', (raw) => {
+    expect(getQuantityError(raw, AVAILABLE)).toBe(QUANTITY_ERRORS.NOT_A_NUMBER)
+  })
```
DavidBarker-EE: **Minor.** Good coverage, and the two precedence tests plus the `max = 0` case are real contracts rather than restatements of the code. A few plausible-bug cases are still unpinned:

- `getQuantityError('11.22', 11.22)` returning null, which defends `>` against a future `>=`. It is currently only implied by one `isValidQuantity` row.
- `'5.'` and `'.'`, both currently accepted or misreported (see the regex comment).
- `'0.00010'`, textually five decimals of a numerically four-decimal value, which pins the string-based precision rule against a future float rewrite.
- `'+5'` and `'+.5'`, asymmetric with the handling of `-`.
- A number-typed `raw` through `getQuantityError`. Only `normaliseQuantityInput` is tested with `11.22`.

src/client/javascripts/land-grants/select-actions-page.test.js
```diff
@@ -1924,3 +1958,233 @@ describe('initSelectActionsPage', () => {
     expect(cmor1Checkbox.disabled).toBe(false)
   })
 })
+
+describe('quantity input validation', () => {
```
DavidBarker-EE: These read as behaviour tests, which is what I want to see: real `input` and `blur` events, rendered output, the `Error:` prefix, the error classes, `aria-describedby` composition, summary hrefs and the submitted payload.

The gaps are the cases behind the two defects flagged in `select-actions-page.js`: focus then blur with no edit on a server-rendered error, which must not clear it, and unticking an errored action, which must remove its summary entry and inline error. Also missing: an unbounded action with no `max` driven through the DOM, and a note that submission still proceeds with an invalid value, since there is no client-side submit blocking and the server is the backstop.

DavidBarker-EE (dismissed):

acceptance/test/features/action-selection.feature
```diff
@@ -33,7 +33,17 @@ Feature: Action Selection
         # RULE: partial action hectares cannot be zero
         When the user enters "0" hectares for action "CSAM3"
         And continues
-        Then the user should see "Enter a quantity for Herbal leys: CSAM3" for action "CSAM3"
+        Then the user should see "Value must be greater than 0" for action "CSAM3"
+
+        # RULE: partial action hectares cannot be negative
```
ScenarioFactory: Yeah you’re extending the tests again. Just fix what needs fixing to keep them green reallg

acceptance/test/steps/action.steps.js
ScenarioFactory: Don’t extend the tests, just do the minimum changes to keep them green.

acceptance/test/features/action-selection.feature
```diff
@@ -46,7 +46,7 @@ Feature: Action Selection
         # RULE: partial action hectares must be 4 decimal places or less
         When the user enters "0.27666" hectares for action "CSAM3"
         And continues
-        Then the user should see "Quantity for Herbal leys: CSAM3 must be 4 decimal places or fewer" for action "CSAM3"
+        Then the user should see "No more than 4 dp" for action "CSAM3"
```
ScenarioFactory: Is that from the prototype? ‘dp’ seems a bit vague

swdpcomputing: I think that might have been a missed merge conflict. Just checking.

src/client/javascripts/land-grants/action-checkbox-state.js
```diff
@@ -0,0 +1,134 @@
...
+ * A non-quantity action's hidden field carrying its chosen area.
+ * @param {HTMLInputElement} checkbox
+ * @returns {HTMLInputElement | null}
+ */
+export function getChosenAreaField(checkbox) {
+  const field = /** @type {HTMLInputElement | null} */ (
+    document.getElementById(getActionQuantityFieldName(checkbox.value))
+  )
+  return field?.type === 'hidden' ? field : null
+}
```
DavidBarker-EE: These two look identical

src/client/javascripts/land-grants/action-checkbox-state.js
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * What a checked action currently holds - its own hidden field (non-quantity)
+ * or data-total-chosen-area (quantity action, which has no hidden field of
+ * its own).
+ * @param {HTMLInputElement} checkbox
+ * @returns {number | undefined}
+ */
+export function getChosenArea(checkbox) {
```
DavidBarker-EE: I'm seeing loads of functions here that kind of all do the same thing. Alot of what comes before this comment can be easily condensed into much less code and a single wrapper function around `HTMLInputElement.getAttribute`?

src/server/land-grants/controllers/select-actions-base-page.controller.js
```diff
@@ -8,10 +8,36 @@ import {
...
+function normaliseQuantityFields(payload) {
+  return Object.fromEntries(
+    Object.entries(payload).map(([key, value]) =>
+      key.startsWith(ACTION_QUANTITY_FIELD_PREFIX) && typeof value === 'string'
+        ? [key, normaliseQuantityInput(value)]
+        : [key, value]
+    )
+  )
+}
+
```
DavidBarker-EE: This is already a pretty large file, much larger than a controller should be. Both of these look quite generic anyway, so maybe in shared?

swdpcomputing: More trimmed some generic parts. But done.

src/client/javascripts/land-grants/action-checkbox-state.js
swdpcomputing: Consolidated.

DavidBarker-EE (approved):

